Issue #2, opened 2022-08-02
Separate collision functions from the C types


itzpr3d4t0r:
We plan to have many collision functions for our shapes, and we will probably end up having a case like this:
```Python
circle = Circle(10, 4, 20)
line = Line((5, 5), (10, 10))

circle.collideline(line)
line.collidecircle(circle)
```
This means we have duplicated code for the same collision function.
I propose we take the collision functions themselves(the ones that calculate the true/false value) and put them in standalone internal C functions in another file like `collisions.c` that we include in `line.c` and `cirlce.c`.

In that way we could have a single collision implementation that will clean up code a lot. We could have something like:
```C
#include "line.h"
#include "circle.h"

int collision_line_circle(pgCircleBase *c, pgLineBase *l);
int collision_line_rect(pgLineBase *c, SDL_Rect *r);
int collision_circle_rect(pgCircleBase *c, SDL_Rect *r);
```

PR #6, opened 2022-08-02
C: added fastcall implementations for `Line` and seperated the collisions

Emc2356:
this should also tackle #2

### DIFF
--- a/src_c/line.c
+++ b/src_c/line.c
@@ -1,6 +1,7 @@
 #include "include/pygame.h"
 #include "include/line.h"
 #include "include/base.h"
+#include "include/collisions.h"
 
 #include <limits.h>
 #include <math.h>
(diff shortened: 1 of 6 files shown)


src_c/line.c
```diff
@@ -1,94 +1,31 @@
 #include "include/pygame.h"
```
itzpr3d4t0r, line 1: Could you change these? the actual file name is "Include" not "include", it's driving my IDE crazy and i don't think it's a good thing. We either rename the "Include" file to "include" or change each `#include` in the .c files

src_c/line.c
```diff
@@ -766,21 +619,18 @@ pg_line_iterator(pgLineObject *self)
     return iter;
 }
 
-#define __LINE_GETSET_NAME(name)                                          \
```
itzpr3d4t0r, line 769: You need to tackle this formatting problem. You see how clean this was with Clang format? you either disable your auto format on save if on VScode or something else, but we can't have a 500 lines change of just formatting, it's making it harder to spot the changes.

src_c/line.c
```diff
@@ -125,270 +59,194 @@ pg_line_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
...
-{
+static int
+pgLine_FromObject(PyObject *obj, pgLineBase* out) {
     Py_ssize_t length;
-    PyObject *fseq = NULL;
+    PyObject* fseq = NULL;
 
     if (pgLine_Check(obj)) {
-        return &((pgLineObject *)obj)->line;
+        *out = ((pgLineObject *)obj)->line;
```
itzpr3d4t0r, line 76: In my opinion you should use a memcpy here, in that way you won't have to copy each attribute when you exit the function, kinda like the following:
```C
memcpy(out, &((pgLineObject *)obj)->line, sizeof(double)*4);
return 1;
```
the __init__ function would then become(for example)

```C
static int
pg_line_init(pgLineObject *self, PyObject *args, PyObject *kwds) {
    pgLineBase argrect;

    if (!pgLine_FromObject(args, &(self->line))) {
        PyErr_SetString(PyExc_TypeError, "Argument must be rect style object");
        return -1;
    }
    return 0;
}
```

PR #10, opened 2022-08-03
C: restructured a part of it and added extra collisions

Emc2356:
made a restructure of the code
we now have `geometry.h` instead of `line.h` & `circle.h` 
that allows us to put in declarations like the structures and some functions so all of the geometry members can use them to do stuff like collisions
also added `Line.collidecircle`


src_c/circle.c
```diff
@@ -62,12 +61,12 @@ _pg_circle_set_radius(PyObject *value, pgCircleBase *circle)
...
 
 static int
-pgCircle_FromObject(PyObject *obj, pgCircleBase *temp)
+pgCircle_FromObject(PyObject *obj, pgCircleBase *out)
 {
     Py_ssize_t length;
 
     if (pgCircle_Check(obj)) {
-        memcpy(temp, &((pgCircleObject *)obj)->circle, sizeof(double) * 4);
+        memcpy(out, &((pgCircleObject *)obj)->circle, sizeof(double) * 4);
```
Starbuck5, line 69: Just do `sizeof(pgCircleBase)`. 4 doubles could be wrong, if the struct's members are aligned to certain widths.

src_c/collisions.c
```diff
@@ -66,10 +71,62 @@ pgIntersection_LineLine(pgLineBase *A, pgLineBase *B, double *X, double *Y)
...
+static int
+pgCollision_CirclePoint(pgCircleBase *circle, double Cx, double Cy)
+{
+    double dx = circle->x - Cx;
+    double dy = circle->y - Cy;
+    return dx * dx + dy * dy <= circle->r * circle->r;
+}
+
 static int
 pgCollision_LineCircle(pgLineBase *line, pgCircleBase *circle)
```
Starbuck5, line 96: I see lots of camelCase in this function, and around :eyes:

Issue #1, opened 2022-08-02
geometry.colliderays() or line.colliderays()


itzpr3d4t0r:
Right now we have a `line.raycast(LineType)` function that calculates collision between two lines.
In many cases when raycasting you'll need to check collision between two line sequences. I propose we add a function that does just that, and in particular is in the form:
```Python
colliderays(seq1: Sequence[LineType], seq2: Sequence[LineType]) -> Sequence[Union[Tuple[float, float], None]]
```
The function will check each line in the first sequence and search for collision with each line in the second sequence. If it didn't find any appends `None` to the return list, else appends the collision point.
We could even make it so you can pass in an optional parameter that lets you choose to make the function return the collision positions or simply  `[] or None `.
We could have it for 2 line sequences but also between one line and a sequence of lines in which case you'd probably want to implement in in `line.c`:
```Python
- geometry.colliderays(seq1: Sequence[LineType], seq2: Sequence[LineType]) -> Sequence[Union[Tuple[float, float], None]]
- line.colliderays(sequence: Sequence[LineType]) -> Union[Tuple[float, float], None]
```

PR #17, opened 2022-08-03
instead of doing one per call we do multiple per Line.raycast call

Emc2356:
#1 complete this


src_c/methods.c
```diff
@@ -0,0 +1,102 @@
...
+    B_len = PySequence_Fast_GET_SIZE(args[1]);
+
+    if (A_len == 0) {
+        return PyList_New(0);
+    }
+    if (B_len == 0) {
+        PyObject *ret = PyList_New(A_len);
+        for (i = 0; i < A_len; i++) {
+            PyList_SET_ITEM(ret, i, Py_None);
+        }
```
itzpr3d4t0r, line 46: This can segfault, `Py_None` is a singleton and needs to be treated line any other object. `PyList_SET_ITEM` steals a reference to the object to be appended, so you are decreasing the Py_None refcount on each append (No good).
instead you should do the following
```suggestion
        for (i = 0; i < A_len; i++) {
            Py_INCREF(Py_None);
            PyList_SET_ITEM(ret, i, Py_None);
        }
```

PR #23, opened 2022-08-04
added `python -m pip install pygame` to run unittests

Emc2356:


itzpr3d4t0r:
Closing this as we already pushed a change with direct commit

PR #34, opened 2022-08-05
Updated circle class and benchmark.

itzpr3d4t0r:
The Circle python class didn't really reflect the current C implementation, so i modified it a bit and remade the circle benchmark suite.
Future functions will be added to the bechmark when we make them.


novialriptide:
Why not get rid of the old benchmarks?

itzpr3d4t0r:
Yep, working on it

PR #38, opened 2022-08-07
Added center, area and circumference attributes.

itzpr3d4t0r:
We didn't have many circle attributes that are commonplace, so i thought that adding those would not only be a necessary addition, but they would also make working with them from python a bit easier.

### DIFF
--- a/src_c/circle.c
+++ b/src_c/circle.c
@@ -7,8 +7,8 @@
 #include <stddef.h>
 #include <math.h>
 
-#define PI (double)3.14159265358979323846264
-#define TAU (double)6.28318530717958647692528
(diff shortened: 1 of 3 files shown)


src_c/circle.c
```diff
@@ -7,6 +7,8 @@
 #include <stddef.h>
 #include <math.h>
 
+#define PI (double)3.14159265358979323846264
+#define TAU (double)6.28318530717958647692528
```
Emc2356, line 11: i would think that you dont need to cast to double here as firstly you dont have float literals and secondly the number is already long enough that it cant be a float

PR #42, opened 2022-08-08
created SIMD version of line-rect intersection

Emc2356:
this version is about 3 times faster than the normal path


setup.py
```diff
@@ -3,8 +3,21 @@
 import shlex
 import glob
 import sys
+import os
 
-extensions = [Extension("geometry", sources=["src_c/geometry.c"])]
+
+compiler_options = {"unix": ["-mavx2"], "msvc": ["/arch:AVX2"]}
```
Starbuck5, line 9: This is not necessary. Since you're using intrinsics, the compiler isn't emitting the symbols, you are. With these options, the project will not run at all on older systems, I believe.

src_c/collisions.c
```diff
@@ -204,6 +192,10 @@ static int
 pgIntersection_LineRect(pgLineBase *line, SDL_Rect *rect, double *X, double *Y,
                         double *T)
 {
+#ifdef __AVX2__
+    return pgIntersection_LineRect_avx2(line, rect, X, Y, T);
```
Starbuck5, line 196: This needs a runtime check as well.

src_c/collisions_avx2.c
```diff
@@ -0,0 +1,90 @@
+#include "include/pygame.h"
+#include "include/geometry.h"
+
+#include <immintrin.h>
+
+
+static PG_FORCEINLINE int
+pgIntersection_LineRect_avx2(pgLineBase *line, SDL_Rect *rect, double *X, double *Y, double *T) {
```
itzpr3d4t0r, line 8: Could you you use better names here? line xintersect and yintersect for example? T as well

src_c/collisions_avx2.c
```diff
@@ -0,0 +1,90 @@
...
+    for (i = 0; i < 4; i++) {
+        double t_ = ((double*)&t_256d)[i];
+        double u_ = ((double*)&u_256d)[i];
+        if (t_ >= 0 && t_ <= 1 && u_ >= 0 && u_ <= 1) {
+           t = MIN(t, t_); 
+        }
+    }
+
+    if (t <= 1 && t >= 0) {
+        if (T) *T = t;
```
itzpr3d4t0r, line 82: this is not readable at all, could you improve variable names here as well?

src_c/collisions_avx2.c
```diff
@@ -0,0 +1,90 @@
...
+    );
+
+    __m256d t_zero_256d = _mm256_cmp_pd(t_256d, _mm256_setzero_pd(), _CMP_GE_OQ);
+    __m256d t_one_256d = _mm256_cmp_pd(t_256d, _mm256_set1_pd(1.0), _CMP_LE_OQ);
+    __m256d u_zero_256d = _mm256_cmp_pd(u_256d, _mm256_setzero_pd(), _CMP_GE_OQ);
+    __m256d u_one_256d = _mm256_cmp_pd(u_256d, _mm256_set1_pd(1.0), _CMP_LE_OQ);
+    __m256d t_u_256d = _mm256_and_pd(
+        _mm256_and_pd(t_zero_256d, t_one_256d),
+        _mm256_and_pd(u_zero_256d, u_one_256d)
+    );
```
itzpr3d4t0r, line 65: Can you add some comments in between these? so that in the future if we wanna change this it will be easier and it will be easier to understand if one knows little about intrinsics

src_c/collisions_avx2.c
```diff
@@ -0,0 +1,100 @@
...
+
+    // we check this condition t >= 0 && t <= 1 && u >= 0 && u <= 1
+    __m256d t_zero_256d =
+        _mm256_cmp_pd(t_256d, _mm256_setzero_pd(), _CMP_GE_OQ);
+    __m256d t_one_256d =
+        _mm256_cmp_pd(t_256d, _mm256_set1_pd(1.0), _CMP_LE_OQ);
+    __m256d u_zero_256d =
+        _mm256_cmp_pd(u_256d, _mm256_setzero_pd(), _CMP_GE_OQ);
+    __m256d u_one_256d =
+        _mm256_cmp_pd(u_256d, _mm256_set1_pd(1.0), _CMP_LE_OQ);
```
itzpr3d4t0r, line 71: just a question, wouldn't it be better to create a single `__m256d zeroes = _mm256_setzero_pd()` and `__m256d ones = _mm256_set1_pd(1.0)` and reuse those in the comparisons instead of creating them twice?

src_c/collisions.c
```diff
@@ -1,26 +1,14 @@
+#ifdef __AVX2__
+#include "collisions_avx2.c"
```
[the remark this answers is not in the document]
Emc2356: i don't think this is necessary as we already have a header for collisions.c and people are not supposed to include this directly, but i would like @Starbuck5's input on this one

PR #43, opened 2022-08-08
Make all shapes use floats

itzpr3d4t0r:
As a first implementation, we decided to use double as number precision for positions and data for shapes. This turned out to be unnecessary and weird, as game developement usually implements stuff with floats. Using floats will enable us to write more efficient functions, both normal and SIMD functions, which will benefit a lot from this change.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,8 +3,21 @@ import subprocess
 import shlex
 import glob
 import sys
+import os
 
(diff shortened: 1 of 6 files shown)


itzpr3d4t0r:
Apparently using floats does not speed things much, but adds overhead when interacting with python, so we are closing this for now.

PR #44, opened 2022-08-14
Add line support to collideswith

itzpr3d4t0r:
The `circle.collideswith()` function didn't have support for Circle-Line collision and had a wrong behaviour for unsupported types, so i added support for the Line type and corrected the wrong behaviour. I also added more benchmarks and the missing collideswith() unittests.


Emc2356 (approved):
LGTM!

PR #45, opened 2022-08-14
Added circle.contains()

itzpr3d4t0r:
This PR adds the `circle.contains()` function to verify that a shape is inside the circle.


novialriptide:
Shouldn't this be renamed to `collidecircle()`?

itzpr3d4t0r:
> Shouldn't this be renamed to `collidecircle()`?

Unlike `collidecircle`, this function returns `True` only when the second circle is completely inside the first. `collidecircle()` returns `True` when the two circles are partially colliding whereas `contains` does not. This function reflects the `rect.contains()` function, that has the same use.

itzpr3d4t0r:
Marking this as a draft as i want this function to work with any shape instead of just circle.

itzpr3d4t0r:
Closing this because there are too many conflicts to resolve.

PR #46, opened 2022-08-15
Add circle.intersect()

itzpr3d4t0r:
We have a line.raycast function that calculates intersection between a line and different shapes, but we don't have a function for circle intersection. This function returns `None` if no intersection or infinite intersections are found; returns A list of two tuples representing the intersection points otherwise.

**We could also make this a general circle.intersect() function that supports any shape for intersection**


itzpr3d4t0r:
I'll open another PR about this that's updated. Closing it for now

PR #47, opened 2022-08-22
implemented line-rect collision with AVX2 and small restructure

Emc2356:
implemented line-rect collision with AVX2 and restructured the structure, duh, in order to make it easier in case we want to implement these functions with other instruction sets

### DIFF
--- a/src_c/collisions.c
+++ b/src_c/collisions.c
@@ -1,13 +1,15 @@
-#ifdef __AVX2__
-#include <immintrin.h>
-#endif
-
 #include "include/collisions.h"
+#include "simd_collisions.h"
 #include <stdio.h>
 
-#ifndef DOT2D
-#define DOT2D(X0, Y0, X1, Y1) ((X0) * (X1) + (Y0) * (Y1))
-#endif /* ~DOT2D */
+#if defined(__GNUC__) && \
+    (__GNUC__ > 2 || (__GNUC__ == 2 && (__GNUC_MINOR__ > 95)))
+#define LIKELY(x) __builtin_expect(!!(x), 1)
+#define UNLIKELY(x) __builtin_expect(!!(x), 0)
+#else
+#define LIKELY(x) (x)
+#define UNLIKELY(x) (x)
+#endif
 
 static int
 pgCollision_LineLine(pgLineBase *A, pgLineBase *B)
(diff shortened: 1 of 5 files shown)


itzpr3d4t0r:
I can't find the actual `line.colliderect()` function tho

src_c/collisions.c
```diff
@@ -323,7 +237,22 @@ pgIntersection_LineRect(pgLineBase *line, SDL_Rect *rect, double *X, double *Y,
...
+    double w = (double)rect->w;
+    double h = (double)rect->h;
+
+    pgLineBase a = {x, y, x + w, y};
+    pgLineBase b = {x, y, x, y + h};
+    pgLineBase c = {x, y + h, x + w, y + h};
+    pgLineBase d = {x + w, y, x + w, y + h};
+
+    return pgCollision_LineLine(line, &a) && pgCollision_LineLine(line, &b) &&
+           pgCollision_LineLine(line, &c) && pgCollision_LineLine(line, &d);
```
itzpr3d4t0r, line 254: Shouldn't these be `||`s?

src_c/simd_collisions_avx2.c
```diff
@@ -0,0 +1,172 @@
...
+
+    // calculate the t values
+    __m256d t_256d = _mm256_sub_pd(_mm256_mul_pd(x1_m_x3_256d, y3_m_y4_256d),
+                                   _mm256_mul_pd(y1_m_y3_256d, x3_m_x4_256d));
+    t_256d = _mm256_div_pd(t_256d, den_256d);
+
+    // calculate the u values
+    __m256d u_256d = _mm256_sub_pd(_mm256_mul_pd(x1_m_x2_256d, y1_m_y3_256d),
+                                   _mm256_mul_pd(y1_m_y2_256d, x1_m_x3_256d));
+    u_256d =
```
Starbuck5, line 156: I have not fully thought this through, but
rather than multiplying this by -1 to negate it, could you instead flip the comparison operators on what you compare this too?

Emc2356: the way to do this is to compare for smaller or equal to 0 and larger or equal to -1, i dont think this gives us any speed improvement as instead of calling one AVX2 SIMD intrinsic we create an extra register with -1.
PS. this idea of just flipping the comparisons wont work cause we will go from `0 <= u <= 1` to `1 <= u <= 0` thus rendering it invalid.

PR #50, opened 2022-08-27
implemented `Line.move` & `Line.move_ip`

Emc2356:


test/test_line.py
```diff
@@ -358,6 +358,44 @@ def test_meth_colliderect(self):
         with self.assertRaises(TypeError):
             A.colliderect(1, 5)
 
+    def test_meth_move(self):
```
Starbuck5, line 361: Should at least one integer be passed as well to test that? It doesn't "just" take floats, presumably.

test/test_line.py
```diff
@@ -358,6 +358,44 @@ def test_meth_colliderect(self):
...
+        self.assertEqual(ret.y2, 6.4)
+
+        with self.assertRaises(TypeError):
+            line.move()
+
+        with self.assertRaises(TypeError):
+            line.move(1)
+
+        with self.assertRaises(TypeError):
+            line.move(1, 2, 3)
```
novialriptide, line 378: Another test I could suggest adding is if the user does `line.move("3", "2")`

test/test_line.py
```diff
@@ -358,6 +358,44 @@ def test_meth_colliderect(self):
...
+        self.assertEqual(line.y2, 6.4)
+
+        with self.assertRaises(TypeError):
+            line.move()
+
+        with self.assertRaises(TypeError):
+            line.move(1)
+
+        with self.assertRaises(TypeError):
+            line.move(1, 2, 3)
```
novialriptide, line 397: As well as here.

novialriptide (approved):
LGTM!

src_c/line.c
```diff
@@ -389,6 +389,72 @@ pg_line_colliderect(pgLineObject *self, PyObject *args)
...
+    }
+
+    pgLineObject *ret =
+        _pg_line_subtype_new4(Py_TYPE(self), self->line.x1, self->line.y1,
+                              self->line.x2, self->line.y2);
+
+    ret->line.x1 += Dx;
+    ret->line.y1 += Dy;
+    ret->line.x2 += Dx;
+    ret->line.y2 += Dy;
```
itzpr3d4t0r, line 419: Can't we do this in a single line? 
If line length is an issue you can always put `pgLineObject *line = &pgLine_AsLine(self)` and use that instead of repeating `self->line.`

maqa41:
hey, you can assign anything to me now.

test/test_line.py
Emc2356: this is a test that already exists in the rect_test.py in pygame as it handles rects, i dont think we need to test that but if @Starbuck5 thinks it is a good idea i would be more than happy to implement it

src_c/line.c
Emc2356: to be honest i prefer it this way as it is a bit more explicit but if you have any other reason other than lines too long i prefer it this way

test/test_line.py
Emc2356: i dont really understand what you want to me to do and i would like you to elaborate.

src_c/line.c
```diff
@@ -389,6 +389,72 @@ pg_line_colliderect(pgLineObject *self, PyObject *args)
...
+
+    return ret;
```
itzpr3d4t0r, line 421: Right now this code has the wrong types: `_pg_line_subtype_new4` returns a `PyObject *`.
Just do this code fragment in a single line please: 
```suggestion
    return pgLine_New4(self->line.x1 + Dx, self->line.y1 + Dy,
                       self->line.x2 + Dx, self->line.y2 + Dy);
```

itzpr3d4t0r (changes requested):

src_c/line.c
itzpr3d4t0r: This will have right types, be more readable and make more sense.

test/test_line.py
```diff
@@ -358,6 +358,44 @@ def test_meth_colliderect(self):
...
+
+        line.move_ip(1, 2)
+
+        self.assertEqual(line.x1, 2.1)
+        self.assertEqual(line.y1, 4.2)
+        self.assertEqual(line.x2, 4.3)
+        self.assertEqual(line.y2, 6.4)
+
+        with self.assertRaises(TypeError):
+            line.move()
```
novialriptide, line 391: This doesn't actually test for `.move_ip()`.

novialriptide (changes requested):
Found an issue, also I'd recommend adding strings as parameters tests.

src_c/line.c
Emc2356: this wouldnt work as we need to return the correct subtype

Emc2356: although something like this is preferable
```py
return _pg_line_subtype_new4(Py_TYPE(self), self->line.x1 + Dx, self->line.y1 + Dy, self->line.x2 + Dx, self->line.y2 + Dy);
```

test/test_line.py
Emc2356: good catch

itzpr3d4t0r (approved):
LGTM!

novialriptide:
I would add a test for invalid types.

PR #53, opened 2022-08-28
Update move(), move_ip() to be same with the line implementation

itzpr3d4t0r:
Made code clearer, more compact and in line with the line implementation (pun not intended).
After discussions on discord we decided to remove the (circle only) functionality to support a single argument in move/move_ip. This is no longer a thing.


Emc2356 (commented):
LGTM!

Emc2356:
i think it is ok to merge this

PR #54, opened 2022-08-28
Add diameter attribute to circle

itzpr3d4t0r:


src_c/circle.c
```diff
@@ -653,6 +673,10 @@ static PyGetSetDef pg_circle_getsets[] = {
     {"r", (getter)pg_circle_getr, (setter)pg_circle_setr, NULL, NULL},
     {"r_sqr", (getter)pg_circle_getr_sqr, (setter)pg_circle_setr_sqr, NULL,
      NULL},
+    {"d", (getter)pg_circle_getdiameter, (setter)pg_circle_setdiameter, NULL,
```
novialriptide, line 676: I'm not sure about the diameter attribute having an alias (`d` and `diameter`). I think this is usually used when we deprecate something. We can keep this in for now, but let's think about removing them in a future PR.

Emc2356, line 676: i dont think we should have an alias for diameter

test/test_circle.py
```diff
@@ -325,6 +325,40 @@ def test_circumference_del(self):
...
+
+        c.r_sqr = 100
+        self.assertEqual(c.diameter, 20.0)
+        self.assertEqual(c.d, 20.0)
+
+    def test_diameter_invalid_value(self):
+        """Ensures the diameter handles invalid values correctly."""
+        c = Circle(0, 0, 1)
+
+        for value in (None, [], "1", (1,), [1, 2, 3], -1, 0):
```
novialriptide, line 351: This'll be funny if we let this in. Haha.

novialriptide (approved):
Everything looks fine so far, I just have a problem with the attributes with alias.

src_c/circle.c
itzpr3d4t0r: I'll answer this the way i answered in discord: 
```
also about the circle.diameter attribute i've aliased that because rect aliases its attributes as well
and could be useful to have both, hurts no one
so if one needs readability will go for "diameter", and if one wants compact code will go for "d" 
as long as the aliases are not confusing i think we are fine with them
```

PR #55, opened 2022-08-29
Create line.rst

tank-king:


docs/line.rst
```diff
@@ -0,0 +1,112 @@
...
+
+        .. ## Line.copy ##
+
+    .. method:: raycast
+
+        | :sl:`cast a ray against a list of shapes/colliders`
+        | :sg:`raycast(List[Union[Line, Rect, Circle]]) -> Union[Tuple[float, float], None]`
+
+        Performs a raycast against a list of supported shapes, and returns whether the line
+        intersects with each of the shapes or not.
```
itzpr3d4t0r, line 48: This kinda makes it soud like this function returns a bool. Could be enough to say `...and returns the closest intersection found, None if there aren't any.`
And right below this you could rephrase to 
`The intersection point is a tuple containing the x and y coordinate of the closest intersection point.`
This way it's simpler, more compact and easier to read.

Another thing to add could be a list of the shapes you can raycast against, that are `Line,` `Circle`, `Rect` and `Polygon`. Also specify that the behaviour of raycast() is always the same, no matter the shape used. The sequence can either be a homogeneous sequence or an heterogeneus one, behaviour will always be the same.

docs/line.rst
```diff
@@ -0,0 +1,112 @@
...
+        .. ## Line.collideline ##
+
+    .. method:: collidepoint
+
+        | :sl:`check if a point is on a line or not`
+        | :sg:`collidepoint(Point) -> bool`
+        | :sg:`collidepoint(x, y) -> bool`
+        | :sg:`collidepoint(PointLike) -> bool`
+
+        Returns True if the point is on the line. Otherwise returns False.
```
itzpr3d4t0r, line 72: A little change, apply this to the other places like this. Also remember to use ` around python stuff like True, False and None.
```suggestion
        Returns `True` if the point is on the line, `False` otherwise.
```

docs/line.rst
```diff
@@ -0,0 +1,112 @@
...
+        Returns True if the line collides with the Rectangle. Otherwise returns False.
+
+        .. ## Line.colliderect ##
+
+    .. method:: as_rect
+
+        | :sl:`returns the rectangle bounding the given line`
+        | :sg:`as_rect() -> Rect`
+
+        Returns a rectangle such that the given line is the diagonal of the rectangle.
```
itzpr3d4t0r, line 100: Maybe just say that it returns the smallest rectangle that encloses the line.

docs/line.rst
```diff
@@ -0,0 +1,112 @@
...
+
+        .. ## Line.as_rect ##
+
+    .. method:: update
+
+        | :sl:`update the current line with new values`
+        | :sg:`update(Line) -> None`
+        | :sg:`update(LineLike) -> None`
+
+        Updates the current line values to the new values provided. Returns None.
```
itzpr3d4t0r, line 110:
```suggestion
        Updates the line endpoints to the new provided positions. This function always returns `None`.
```

itzpr3d4t0r:
This is no longer needed as a V1 docs for everything has already been pushed

PR #63, opened 2022-09-12
Add another circle constructor

novialriptide:
Fixes: https://github.com/novialriptide/pygame_geometry/issues/62


itzpr3d4t0r:
You should do the same but for the other branch in the function, the one with `PySequence_Check()`

novialriptide:
> You should do the same but for the other branch in the function, the one with `PySequence_Check()`

What do you mean by "other branch in the function"?

itzpr3d4t0r:
<https://github.com/novialriptide/pygame_geometry/pull/63/files#diff-ac4e813f079482c6454db108a4a2650020afeab43e9a6faec6ea13c9416af989R111-R137> add an `else if (length == 2)` in between those

itzpr3d4t0r:
Past line 111 basically

Emc2356 (approved):
LGTM!
i really like the idea of having a constructor with the coordinates and then the radius.

itzpr3d4t0r:
We should also add that 2 arguments handling to the fastcall version of `pgCircle_FromObject`: `pgCircle_FromObjectFastcall`

PR #67, opened 2022-09-14
Applied PEP8 naming convention to object methods

itzpr3d4t0r:
This PR aims to rename many methods following the PEP8 convention, this applies to:
1. `collidepoint` -> `collide_point`
2. `colliderect` -> `collide_rect`
3. `collidecircle` -> `collide_circle`
4. `collideline` -> `collide_line`
5. `collideswith` -> `collides_with`
6. ~~`raycast` -> `ray_cast `( this one i'm not so sure about)~~ **no longer a thing**

I've also slightly reorganised the line benchmark suite to match the circle one and made some simplifications in order to make files more compact.


novialriptide:
`raycast` should stay the same.

Emc2356:
for now raycast should remain as raycast
BUT i am thinking of renaming raycast to lets say `Line.intersect` and make a new raycast function that makes the Line act like a ray and basically make it have infinite reach

novialriptide:
Actually, this PR needs to be on hold rn, there are already existing pygame functions like `collidepoint` and `collidepoint`.

novialriptide:
Unfortunately, I'm going to have to close this PR since if this does get merged, the new function names will be inconsistent with the existing pygame functions. Let me know if you have other opinions.

PR #72, opened 2022-09-16
deleted old unsupported python implementations.

itzpr3d4t0r:
removed unused files.
Converted old circle examples to geometry shapes.


novialriptide:
LGTM from a quick skim.

PR #73, opened 2022-09-16
implemented `geometry.regular_polygon`

Emc2356:


geometry.pyi
```diff
@@ -180,4 +180,8 @@ class Polygon:
     @overload
     def __init__(self, polygon: Polygon) -> None: ...
     def __copy__(self) -> "Polygon": ...
+    @staticmethod
+    def normal_polygon(
+        sides: int, center: Sequence[float], radius: float, angle: float = 0
```
itzpr3d4t0r, line 185: HEre we should add a `, /` to signify all parameters are positional
```suggestion
        sides: int, center: Sequence[float], radius: float, angle: float = 0, /
```

src_c/polygon.c
```diff
@@ -315,6 +319,73 @@ pg_polygon_dealloc(pgPolygonObject *self)
...
+        if (sides < 0) {
+            return RAISE(PyExc_ValueError,
+                         "the sides can not be a negative number");
+        }
+        return RAISE(PyExc_ValueError, "polygons need at least 3 sides");
+    }
+
+    double *verticies = PyMem_New(double, sides * 2);
+    if (!verticies) {
+        return PyErr_NoMemory();
```
itzpr3d4t0r, line 370: just return `NULL`, errors are already set if there's no memory

src_c/polygon.c
```diff
@@ -315,6 +319,73 @@ pg_polygon_dealloc(pgPolygonObject *self)
...
+            goto error;
+        }
+        sides = (int)PyLong_AsLong(args[0]);
+
+        if (!pg_TwoDoublesFromObj(args[1], &Cx, &Cy) ||
+            !pg_DoubleFromObj(args[2], &radius) ||
+            !pg_DoubleFromObj(args[3], &angle)) {
+            goto error;
+        }
+    }
```
itzpr3d4t0r, line 353: This code has a problem with `PyObject->int` conversion as `PyLong_AsLong` can fail, plus this code can be shaved off of some duplication:
```suggestion
  if (nargs < 3 || nargs > 4) {
      goto error;
  }
  if (!PyLong_Check(args[0])) {
        goto error;
  }
  sides = PyLong_AsLong(args[0]);
  if(PyErr_Occurred()) {
      return NULL;
  }
  if (!pg_TwoDoublesFromObj(args[1], &Cx, &Cy) ||
      !pg_DoubleFromObj(args[2], &radius)) {
      goto error;
  }
  if (nargs == 4 && !pg_DoubleFromObj(args[3], &angle) ) {
      goto error;
  }
```

src_c/polygon.c
```diff
@@ -315,6 +319,73 @@ pg_polygon_dealloc(pgPolygonObject *self)
...
+
+    double *verticies = PyMem_New(double, sides * 2);
+    if (!verticies) {
+        return PyErr_NoMemory();
+    }
+
+    int loop;
+    for (loop = 0; loop < sides; loop++) {
+        verticies[loop * 2 + 0] =
```
itzpr3d4t0r, line 375: why are you adding 0? loop starts from 0 so it always indexes 0 at the start.

src_c/polygon.c
```diff
@@ -315,6 +319,73 @@ pg_polygon_dealloc(pgPolygonObject *self)
...
+            Cx + radius * cos(angle + PI * 2 * loop / sides);
+        verticies[loop * 2 + 1] =
+            Cy + radius * sin(angle + PI * 2 * loop / sides);
```
itzpr3d4t0r, line 378: You are evaluating `PI * 2 * loop / sides` for each vertex. I suggest you calculate `PI * 2 / sides` once and put it in a "fac" variable and use that

src_c/polygon.c
```diff
@@ -315,6 +319,73 @@ pg_polygon_dealloc(pgPolygonObject *self)
...
+    }
+
+    PyObject *ret = pgPolygon_New2(verticies, sides);
+    PyMem_Free(verticies);
+
+    return ret;
+error:
+    return RAISE(PyExc_TypeError, "mismatched argument types");
```
itzpr3d4t0r, line 386: I would specify the required arguments number in this error, so:
```suggestion
    return RAISE(PyExc_TypeError, "mismatched argument types, must either be 3 or 4");
```

src_c/polygon.c
```diff
@@ -342,6 +413,8 @@ pg_polygon_copy(pgPolygonObject *self, PyObject *_null)
 }
 
 static struct PyMethodDef pg_polygon_methods[] = {
+    {"normal_polygon", (PyCFunction)pg_polygon_normal_polygon,
```
itzpr3d4t0r, line 416: Do we really want this to be part of Polygon? couldn't it be a module method? like `geometry.normal_polygon`? Just a question.

itzpr3d4t0r (changes requested):

src_c/polygon.c
itzpr3d4t0r: Ok no this one i'm wrong. But still you should probably do 
```C
return RAISE(PyExc_MemoryError,
                     "cannot allocate memory for the polygon vertices");
```

itzpr3d4t0r: so you get a precise error message instead of an empty memory error

src_c/polygon.c
Emc2356: it has to do with the object that is why it is a static method

geometry.pyi
Emc2356: python 3.8 feature so no

src_c/polygon.c
```diff
@@ -315,6 +319,66 @@ pg_polygon_dealloc(pgPolygonObject *self)
...
+        !pg_DoubleFromObj(args[2], &radius)) {
+        goto error;
+    }
+    if (nargs == 4) {
+        if (!pg_DoubleFromObj(args[3], &angle)) {
+            goto error;
+        }
+    }
+
+    angle *= PI / 180.0;
```
itzpr3d4t0r, line 352: the angle calculation is only useful when the angle is passed as a parameter, you should move that into the if statement:
```suggestion
    if (nargs == 4) {
        if (!pg_DoubleFromObj(args[3], &angle)) {
            goto error;
        }
        angle *= PI / 180.0;
    }

    
```

src_c/polygon.c
```diff
@@ -315,6 +319,66 @@ pg_polygon_dealloc(pgPolygonObject *self)
...
+        if (sides < 0) {
+            return RAISE(PyExc_ValueError,
+                         "the sides can not be a negative number");
+        }
+        return RAISE(PyExc_ValueError, "polygons need at least 3 sides");
+    }
+
+    double *verticies = PyMem_New(double, sides * 2);
+    if (!verticies) {
+        return NULL;
```
itzpr3d4t0r, line 364: I renew my old review on this one, pygame uses RAISE(PyExc_MemoryError..., not NULL, and it can be helpful in tight memory situations, so we should align with them.

src_c/polygon.c
```diff
@@ -315,6 +319,66 @@ pg_polygon_dealloc(pgPolygonObject *self)
...
+
+    if (sides < 3) {
+        if (sides < 0) {
+            return RAISE(PyExc_ValueError,
+                         "the sides can not be a negative number");
+        }
+        return RAISE(PyExc_ValueError, "polygons need at least 3 sides");
+    }
+
+    double *verticies = PyMem_New(double, sides * 2);
```
itzpr3d4t0r, line 362: Mispelled `vertices` with `verticies`:
```suggestion
    double *vertices = PyMem_New(double, sides * 2);
```

src_c/polygon.c
```diff
@@ -315,6 +319,66 @@ pg_polygon_dealloc(pgPolygonObject *self)
...
+    if (!verticies) {
+        return NULL;
+    }
+
+    int loop;
+    for (loop = 0; loop < sides; loop++) {
+        double ang = angle + PI * 2 * loop / sides;
+        verticies[loop * 2] = Cx + radius * cos(ang);
+        verticies[loop * 2 + 1] = Cy + radius * sin(ang);
+    }
```
itzpr3d4t0r, line 372: There are loop invariants that are still calculated for each vertex, just extract the invariant from the loop:
```suggestion
    double fac = PI * 2 / sides;
    for (loop = 0; loop < sides; loop++) {
        double ang = angle + fac  * loop;
        verticies[loop * 2] = Cx + radius * cos(ang);
        verticies[loop * 2 + 1] = Cy + radius * sin(ang);
    }
```

src_c/polygon.c
itzpr3d4t0r: This one is still up

src_c/polygon.c
Emc2356: are you fucking kidding me

src_c/polygon.c
Emc2356: already handled, the wrong number of arguments is caught early in the function

itzpr3d4t0r (approved):
LGTM, didn't check the tests tho, so i'll leave it to someone else

novialriptide:
I'm unsure about this PR because of its API design. Should we create a new constructor for `Polygon` instead of using a static method?

itzpr3d4t0r:
Let's pause a sec on this one, might be wise to implement this as a constructor

src_c/geometry.c
```diff
@@ -8,7 +8,77 @@
...
+                     "invalid number of arguments, expected 3 or 4 arguments");
+    }
+    if (!PyLong_Check(args[0])) {
+        return RAISE(PyExc_TypeError,
+                     "the first parameter must be an integer");
+    }
+    sides = PyLong_AsLong(args[0]);
+    if (sides == -1 && PyErr_Occurred()) {
+        return NULL;
+    }
```
itzpr3d4t0r, line 31: I think it's enough to do this because if `PyLong_AsLong` fails an error is automatically set and is the correct one anyway.
```suggestion
    sides = PyLong_AsLong(args[0]);
    if (PyErr_Occurred()) {
        return NULL;
    }
```

src_c/geometry.c
```diff
@@ -8,7 +8,77 @@
...
+        angle *= PI / 180.0;
+    }
+
+    if (sides < 3) {
+        if (sides < 0) {
+            return RAISE(PyExc_ValueError,
+                         "the sides can not be a negative number");
+        }
+        return RAISE(PyExc_ValueError, "polygons need at least 3 sides");
+    }
```
itzpr3d4t0r, line 55: This should be just after where the sides variable gets assigned

src_c/geometry.c
```diff
@@ -8,7 +8,77 @@
...
+
+    double *vertices = PyMem_New(double, sides * 2);
+    if (!vertices) {
+        return RAISE(PyExc_MemoryError,
+                     "cannot allocate memory for the polygon vertices");
+    }
+
+    int loop;
+    double fac = PI * 2 / sides;
```
itzpr3d4t0r, line 64: PI * 2 should be defined as a macro and not calculated on each call
```suggestion
    double fac = TWOPI / sides;
```

src_c/geometry.c
itzpr3d4t0r: Plus you can just check for PyErr_Occurred() because the error is always set when it fails, but it can happen that PyLong_AsLong returns -1 but it was intended to return -1.

PR #74, opened 2022-09-17
Support black color in `get_new_circle_surf`

blankRiot96:
Function did not allow black circles as it would `set_colorkey` on it. This way the surface starts off transparent.


itzpr3d4t0r:
Thanks for contributing! We will need better examples tho, i wrote these in a hurry so didn't mind quality or **most importantly readability**.

PR #75, opened 2022-09-19
Example script using normal_polygon on class Polygon

newpaxonian:
This is a simple example script using the new function normal_polygon, previously commented on https://discord.com/channels/772505616680878080/1002836330482511943/1021180402645336175

### DIFF
--- a/examples/regular_polygons.py
+++ b/examples/regular_polygons.py
@@ -1,0 +1,89 @@
+import random
+import sys
+
+import geometry
+import pygame as pg
+
+FPS = 60
+RES = WIDTH, HEIGHT = 640, 480
+DARK_GREY = (10, 10, 10)
+SIDES = [3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
+COLORS = {
+    3: (255, 0, 0),
+    4: (0, 255, 0),
+    5: (0, 0, 255),
+    6: (255, 255, 0),
+    7: (0, 255, 255),
+    8: (255, 0, 255),
+    9: (255, 255, 255),
+    10: (255, 127, 0),
+    11: (127, 0, 255),
+    12: (0, 255, 127),
+}
+
+
+class Game:
+    def __init__(self):
+        pg.init()
+        # pg.mouse.set_visible(False)
+        self.display = pg.display
+        self.screen = self.display.set_mode(RES)
+        self.clock = pg.time.Clock()
+        self.colors = COLORS
+        self.radius = 64
+        self.toggle_mouse_button = False
+        self.polygons = []
+
+    def events(self):
+        for event in pg.event.get():
+            if event.type == pg.QUIT or (event.type == pg.KEYDOWN and event.key == pg.K_ESCAPE):
+                print("\nQuitting program...")
+                pg.quit()
+                sys.exit()
+            if pg.mouse.get_pressed()[0]:
+                if self.toggle_mouse_button:
+                    self.toggle_mouse_button = False
+                else:
+                    self.toggle_mouse_button = True
+
+    def draw(self, num_sides=3, color=(255, 255, 255)):
+        box_side = 2 * self.radius
+        box = pg.Surface((box_side, box_side))
+        box.fill(DARK_GREY)
+
+        box_x, box_y = pg.mouse.get_pos()
+
+        polygon_radius = box_side // 2
+        polygon_center = (box_side // 2, box_side // 2)
+
+        polygon = geometry.Polygon.normal_polygon(num_sides, polygon_center, polygon_radius, 30)
+
+        self.screen.fill(DARK_GREY)
+        pg.draw.polygon(box, color, polygon.vertices)
+        self.screen.blit(box, (box_x - box_side // 2, box_y - box_side // 2))
+
+    def update(self):
+        self.display.update()
+
+    def run(self):
+        while True:
+            self.clock.tick(FPS)
+            self.display.set_caption(f"Regular Polygons - FPS {self.clock.get_fps():.1f}")
+
+            self.events()
+
+            if self.toggle_mouse_button:
+                num_sides = SIDES[random.randint(0, len(self.colors) - 1)]
+                color = self.colors[num_sides]
+                self.draw(num_sides=num_sides, color=color)
+
+            self.update()
+
+
+def main():
+    game = Game()
+    game.run()
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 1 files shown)


novialriptide:
Looks like you need to lint your code, you can also ignore the MacOS workflow.

examples/regular_polygons.py
```diff
@@ -0,0 +1,89 @@
...
+
+import geometry
+import pygame as pg
+
+FPS = 60
+RES = WIDTH, HEIGHT = 640, 480
+DARK_GREY = (10, 10, 10)
+SIDES = [3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
+COLORS = {
+    3: (255, 0, 0),
```
novialriptide, line 12: Instead of having this, I think it would be better to just come up with a random RGB and side values instead of having this list here.

examples/regular_polygons.py
```diff
@@ -0,0 +1,89 @@
...
+        self.screen.fill(DARK_GREY)
+        pg.draw.polygon(box, color, polygon.vertices)
+        self.screen.blit(box, (box_x - box_side // 2, box_y - box_side // 2))
+
+    def update(self):
+        self.display.update()
+
+    def run(self):
+        while True:
+            self.clock.tick(FPS)
```
novialriptide, line 70: You could remove the `FPS` variable and change it to `self.clock.tick(60)`

newpaxonian:
All right. I'll review it.

examples/regular_polygons.py
```diff
@@ -0,0 +1,89 @@
...
+        self.toggle_mouse_button = False
+        self.polygons = []
+
+    def events(self):
+        for event in pg.event.get():
+            if event.type == pg.QUIT or (event.type == pg.KEYDOWN and event.key == pg.K_ESCAPE):
+                print("\nQuitting program...")
+                pg.quit()
+                sys.exit()
+            if pg.mouse.get_pressed()[0]:
```
novialriptide, line 43: There's another way to do this, try using `pygame.MOUSEBUTTONDOWN` instead.

examples/regular_polygons.py
```diff
@@ -0,0 +1,89 @@
...
+    12: (0, 255, 127),
+}
+
+
+class Game:
+    def __init__(self):
+        pg.init()
+        # pg.mouse.set_visible(False)
+        self.display = pg.display
+        self.screen = self.display.set_mode(RES)
```
novialriptide, line 30: You could also combine these variables together deleting the variable creation of `RES`.

examples/regular_polygons.py
```diff
@@ -0,0 +1,89 @@
...
+    10: (255, 127, 0),
+    11: (127, 0, 255),
+    12: (0, 255, 127),
+}
+
+
+class Game:
+    def __init__(self):
+        pg.init()
+        # pg.mouse.set_visible(False)
```
novialriptide, line 28: Remove this comment

examples/regular_polygons.py
```diff
@@ -0,0 +1,89 @@
...
+            self.display.set_caption(f"Regular Polygons - FPS {self.clock.get_fps():.1f}")
+
+            self.events()
+
+            if self.toggle_mouse_button:
+                num_sides = SIDES[random.randint(0, len(self.colors) - 1)]
+                color = self.colors[num_sides]
+                self.draw(num_sides=num_sides, color=color)
+
+            self.update()
```
novialriptide, line 80: You can delete this function and move its contents here.

novialriptide (changes requested):

novialriptide:
Great PR! This'll be a nice addition to the examples directory, but I have some change requests.

examples/regular_polygons.py
newpaxonian: ok

examples/regular_polygons.py
newpaxonian: The problem with random colors is that sometimes colors are not that good on the screen.

examples/regular_polygons.py
newpaxonian: I didn't know that. I'll use that variable

examples/regular_polygons.py
newpaxonian: ok

examples/regular_polygons.py
newpaxonian: ok

examples/regular_polygons.py
newpaxonian: Using `pygame.MOUSEBUTTONDOWN` sometimes has glitches when you click the screen. I added a second condition ` if pg.MOUSEBUTTONDOWN and any(pg.mouse.get_pressed())`

examples/regular_polygons.py
```diff
@@ -0,0 +1,106 @@
...
+        while True:
+            self.clock.tick(60)
+            self.display.set_caption(
+                f"Regular Polygons - FPS {self.clock.get_fps():.1f}"
+            )
+
+            self.events()
+
+            if self.toggle_mouse_button:
+                num_sides = SIDES[random.randint(0, len(SIDES) - 1)]
```
novialriptide, line 86: You could use `random.choice()` instead here.

examples/regular_polygons.py
```diff
@@ -0,0 +1,106 @@
...
+        Method events: handles the events
+        """
+        for event in pg.event.get():
+            if event.type == pg.QUIT or \
+                    (event.type == pg.KEYDOWN and event.key == pg.K_ESCAPE):
+                print("\nQuitting program...")
+                pg.quit()
+                sys.exit()
+            if pg.MOUSEBUTTONDOWN and any(pg.mouse.get_pressed()):
+                if self.toggle_mouse_button:
```
novialriptide, line 44: You can condense these four lines to just `self.toggle_mouse_button = not self.toggle_mouse_button`

novialriptide (changes requested):
A few more things

examples/regular_polygons.py
novialriptide: I'll take a further look at this in the near future, for now I'll keep this as unresolved.

examples/regular_polygons.py
newpaxonian: ok, makes sense

examples/regular_polygons.py
newpaxonian: ok. I'll change it

newpaxonian:
I'm correcting those lint issues in the next commit. Thanks for your patience

newpaxonian:
Thanks for all the feedback and corrections. This was my first time collaborating on an open project. Sorry for any inconvenience I may have created...

PR #81, opened 2022-09-20
Polygon functions

maqa41:
Some collision functions added, other useful functions are also added. And also intersection functions are added but they are not bound yet.


novialriptide:
As discussed on Discord, please separate these functions into their own pull request.

PR #83, opened 2022-09-22
line.length attribute. 

ScriptLineStudios:
#79


itzpr3d4t0r:
Hello and thanks for contributing! Now that you opened this PR i think there's more room for changes regarding distance calculations. You should probably modify the existing (C) functions for calculations from what we have now to:
```C
static inline double
pgLine_Length(pgLineBase *line)
{
    double dx = line->x2 - line->x1;
    double dy = line->y2 - line->y1;
    return sqrt(dx * dx + dy * dy);
}

static inline double
pgLine_LengthSquared(pgLineBase *line)
{
    double dx = line->x2 - line->x1;
    double dy = line->y2 - line->y1;
    return dx * dx + dy * dy;
}
```
These will basically address two issues we have right now that are:

- They copy the entire line struct for each call
- They calculate `dx` and `dy` twice

Plus the inline seems fitting here.

src_c/line.c
```diff
@@ -921,6 +924,20 @@ pg_line_setb(pgLineObject *self, PyObject *value, void *closure)
...
+    pgLineBase line;
+    PyObject *line_length;
+    if (!pgLine_FromObject((PyObject*)self, &line)) {
+        Py_DECREF(line_length);
+        return NULL;
+    }
+    double length = pgLine_Length(line);
+    line_length = PyFloat_FromDouble(length);
+    return line_length;
+}
```
itzpr3d4t0r, line 939: Sorry to say this but the function should really just be 
```C
static PyObject *
pg_line_getlength(pgLineObject *self, void *closure)
{
    return PyFloat_FromDouble(pgLine_Length(self->line));
}
```
Hope i can help you learning something useful by explaining what you did wrong and how you can do it better.

So starting from the very top your function is missing the `void *closure` parameter (this is a must have for getters and don't really know what purpose it serves).

The part where you try to extract a line struct from the `self` parameter:
```C
    pgLineBase line;
    PyObject *line_length;
    if (!pgLine_FromObject((PyObject*)self, &line)) {
        Py_DECREF(line_length);
        return NULL;
    }
```
It isn't needed as the `self` parameter is of type `pgLineObject` that already contains a  `pgLineBase` struct called `line`, so to access that it's enough to do:
```
self->line
```

Another thing about the use of `pgLine_FromObject` is that it doesn't set any error so instead of returning `NULL` on failure, you should set an error with `RAISE` and return that. `RAISE` basically sets an error of the type you want, with the error message you want and returns `NULL`.
But all of this isn't needed still as i'll show you in the next section.

---
Now this code fragment:
```C
    double length = pgLine_Length(line);
    line_length = PyFloat_FromDouble(length);
    return line_length;
```

Now we know that line can simply be obtained by doing `self->line`. We could also say that creating a `double length` variable just before the conversion to a `PyObject *` is superflous as the `pgLine_Length` function is already clear enough in what it returns that it renders another variable redundant.

What we should have now is (we now want the address of the struct and not the struct itself so we add `&`):

```C
    line_length = PyFloat_FromDouble(pgLine_Length(&self->line));
    return line_length;
```
We can simplify this even further by exploiting the fact that  `PyFloat_FromDouble` can fail to convert the input, and when it does it sets an error and returns `NULL`, so we can simply return it alltogether.

```C
    return PyFloat_FromDouble(pgLine_Length(&self->line));
```

Hope this was useful and thanks for contributing again!

itzpr3d4t0r:
Don't miond MacOS failing, we are still investigating the cause

src_c/line.c
ScriptLineStudios: Thanks for the detailed explanation. Didn't realize pgLineObject already contained the pgLineBase struct. Thanks again

test/test_line.py
```diff
@@ -285,6 +285,13 @@ def test_attrib_b(self):
         with self.assertRaises(AttributeError):
             del line.b
 
+    def test_attrib_length(self):
+        """a full test for the length attribute"""
+        expected_length = 3.0
+        line = Line(1, 4, 4, 4)
+
+        self.assertEqual(line.length, expected_length)
```
itzpr3d4t0r, line 293: Could you add more tests? like make two different lines with the same length and check if the two are the same, measure a line, move it and remeasure it to check if the length is correctly updated etc.

itzpr3d4t0r:
Also a quick question, do you think you'll be using this library in one of your projects? Do you find these things useful? Thanks.

itzpr3d4t0r:
Forgot to tell you that you still need to add the length attribute in the geometry.pyi file

Emc2356 (commented):
this PR looks good to me, and than for updating `pgLine_Length` to take a pointer, but as predetor said we need more tests

novialriptide (approved):
Hi @ScriptLineStudios! Thanks for contributing to pygame_geometry. In the future, I recommend making a new branch in your forked repository and making your new changes there.

PR #84, opened 2022-09-22
Add new functions for safe and fast tuple generation

itzpr3d4t0r:
This PR adds 3 new C api functions for creating tuples representing points. It also implements part of these functions to existing API (specifically the line one) to make them safer and more compact.

## Added
- **pg_TupleFromDoubles**
    Takes two doubles and returns a tuple containing two PyFloat objects on success, NULL and sets an error otherwise

- **pg_TupleFromInts** 
    Same as the last but for ints

- **pg_PointList_FromArrayDouble**
    Takes an even length double array and returns a list of points: 
`C_arr[1, 2, 3, 4, 5, 6, 7, 8] -> List((1, 2), (3, 4), (5, 6), (7, 8))`

- **pg_PointTuple_FromArrayDouble**
    Takes an even length double array and returns a tupleof points: 
`C_arr[1, 2, 3, 4, 5, 6, 7, 8] -> Tuple((1, 2), (3, 4), (5, 6), (7, 8))`

## Modified
- line `a` getter
    The getter was unsafe as it directly set the tuple items without checking that the conversion `double->PyObject *` was successful. Now with the new functions we have correct checks.

- line `b` getter
    The getter was unsafe as it directly set the tuple items without checking that the conversion `double->PyObject *` was successful. Now with the new functions we have correct checks.

- circle `center` attribute
    The getter was utilizing `Py_BuildValue`, which is slower than the new implementation while still being safe with errors.

- line `raycast` function
    Is now slightly faster because of the new function ( Demonstrated that these functions are faster than `Py_BuildValue` in an already merged pygame PR)

### DIFF
--- a/src_c/include/base.h
+++ b/src_c/include/base.h
@@ -188,4 +188,118 @@ pg_UintFromObjIndex(PyObject *obj, int _index, Uint32 *val)
     return result;
 }
 
+// these return PyObject * on success and NULL on failure.
+
+static PG_FORCE_INLINE PyObject *
+pg_TupleFromDoublePair(double val1, double val2)
+{
+    /*this is demonstrated to be faster than Py_BuildValue*/
+    PyObject *tuple = PyTuple_New(2);
+    if (!tuple)
+        return NULL;
+
+    PyObject *tmp = PyFloat_FromDouble(val1);
+    if (!tmp) {
+        Py_DECREF(tuple);
+        return NULL;
+    }
+    PyTuple_SET_ITEM(tuple, 0, tmp);
+
+    tmp = PyFloat_FromDouble(val2);
+    if (!tmp) {
+        Py_DECREF(tuple);
+        return NULL;
+    }
+    PyTuple_SET_ITEM(tuple, 1, tmp);
+
+    return tuple;
+}
+
+static PG_FORCE_INLINE PyObject *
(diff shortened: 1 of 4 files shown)


src_c/include/base.h
```diff
@@ -188,4 +188,97 @@ pg_UintFromObjIndex(PyObject *obj, int _index, Uint32 *val)
...
+        Py_DECREF(tuple);
+        return NULL;
+    }
+    PyTuple_SET_ITEM(tuple, 1, tmp);
+
+    return tuple;
+}
+
+static PG_FORCE_INLINE PyObject *
+pg_PointSequence_FromArrayDouble(double *array, int len, int ret_list)
```
Emc2356, line 244: the name in the function is ok with me, but maybe change the signature a bit to something like this:
`pg_PointSequence_FromArrayDouble(double *array, int length, int tuple_size, int use_tuple)`

Emc2356 (commented):
i really really like this idea, but i would like to firstly have a number in the function names where it indicates the size of the tuple and in the sequence one to add one extra parameter to indicate the size of the tuple

src_c/include/base.h
itzpr3d4t0r: What should `int tuple_size` do?

Emc2356 (commented):
LGTM!

PR #85, opened 2022-09-22
Reorganized benchmarks

itzpr3d4t0r:
I separated the line and circle `attribute benchmarks` in two `attribute-getters` and `attribute-setters` benchmarks.
Compacted all the objects into the GLOBS dict.
Another important change is making the raycast test not use a sequence of random lines but rather always the same sequence of lines. Without this change performance wasn't correctly measured as it could change heavily depending on the lines generated.


Emc2356:
i do not understand most of this but LGTM!

PR #86, opened 2022-09-22
Core polygon functions

maqa41:


src_c/polygon.c
```diff
@@ -358,7 +467,68 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+        return -1;
+    }
+    self->polygon.c_y = val;
+    return 0;
+}
+
+static PyObject *
+pg_polygon_get_center(pgPolygonObject *self, void *closure)
+{
+    return Py_BuildValue("(dd)", self->polygon.c_x, self->polygon.c_y);
```
itzpr3d4t0r, line 511: You should use the `pg_tuple_from_values_double` function instead of `Py_BuildValue`, it's 40% faster.
```suggestion
    return pg_tuple_from_values_double(self->polygon.c_x, self->polygon.c_y);
```

src_c/polygon.c
```diff
@@ -121,6 +140,8 @@ pgPolygon_FromObject(PyObject *obj, pgPolygonBase *out)
                     return 0;
                 }
             }
+            /*Function to set polygon center*/
+            set_polygon_center_coords(out);
```
itzpr3d4t0r, line 144: Right above this line we have a for loop that runs through the sequence to get the vertices positions. Might be wise to find a different, more efficient strategy, such as repeating that function exploiting the for loop. summing the x and the y.

src_c/polygon.c
```diff
@@ -341,7 +367,90 @@ pg_polygon_copy(pgPolygonObject *self, PyObject *_null)
...
+            if (!pg_DoubleFromObj(args[0], &x) ||
+                !pg_DoubleFromObj(args[1], &y)) {
+                return RAISE(PyExc_TypeError,
+                             "Invalid arguments, must be 2 numbers");
+            }
+            break;
+        default:
+            return RAISE(PyExc_TypeError,
+                         "Polygon.move requires 1 or 2 numbers");
+    }
```
itzpr3d4t0r, line 397: This whole switch i suppose it's derived from my circle move function, but we got rid of that now. and we do not support passing a single value to move functions, you can just pass them a piar of values. Try to rewrite these like the ones we have now.

src_c/polygon.c
```diff
@@ -341,7 +367,90 @@ pg_polygon_copy(pgPolygonObject *self, PyObject *_null)
...
+
+    Py_ssize_t it;
+    for (it = 0; it < self->polygon.verts_num; it++) {
+        self->polygon.vertices[it * 2] += x;
+        self->polygon.vertices[it * 2 + 1] += y;
+    }
+    self->polygon.c_x += x;
+    self->polygon.c_y += y;
+
+    return pgPolygon_New(&(self->polygon));
```
itzpr3d4t0r, line 407: You should use `_pg_polygon_subtype_new2` here
```suggestion
    return _pg_polygon_subtype_new2(Py_TYPE(self), &(self->polygon.vertices), self->polygon.verts_num);
```

src_c/polygon.c
```diff
@@ -341,7 +367,90 @@ pg_polygon_copy(pgPolygonObject *self, PyObject *_null)
...
+            if (!pg_DoubleFromObj(args[0], &x) ||
+                !pg_DoubleFromObj(args[1], &y)) {
+                return RAISE(PyExc_TypeError,
+                             "Invalid arguments, must be 2 numbers");
+            }
+            break;
+        default:
+            return RAISE(PyExc_TypeError,
+                         "Polygon.move_ip requires 1 or 2 numbers");
+    }
```
itzpr3d4t0r, line 438: Same as discussed for the move function here.

src_c/polygon.c
```diff
@@ -358,7 +467,68 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+static int
+pg_polygon_set_center_x(pgPolygonObject *self, PyObject *value, void *closure)
+{
+    double val;
+    DEL_ATTR_NOT_SUPPORTED_CHECK_NO_NAME(value);
+    if (!pg_DoubleFromObj(value, &val)) {
+        PyErr_SetString(PyExc_TypeError, "Expected a number");
+        return -1;
+    }
+    self->polygon.c_x = val;
```
itzpr3d4t0r, line 485: When setting a new center value of a polygon you should move all vertices accordignly, that is missing here.

src_c/polygon.c
```diff
@@ -358,7 +467,68 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+    return 0;
+}
+
+static PyObject *
+pg_polygon_get_center_y(pgPolygonObject *self, void *closure)
+{
+    return PyFloat_FromDouble((int)self->polygon.c_y);
```
itzpr3d4t0r, line 492:
```suggestion
    return PyFloat_FromDouble(self->polygon.c_y);
```

itzpr3d4t0r: PyFloat_FromDouble required a double as input, you are passing int

src_c/polygon.c
```diff
@@ -358,7 +467,68 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+static int
+pg_polygon_set_center_y(pgPolygonObject *self, PyObject *value, void *closure)
+{
+    double val;
+    DEL_ATTR_NOT_SUPPORTED_CHECK_NO_NAME(value);
+    if (!pg_DoubleFromObj(value, &val)) {
+        PyErr_SetString(PyExc_TypeError, "Expected a number");
+        return -1;
+    }
+    self->polygon.c_y = val;
```
itzpr3d4t0r, line 504: When setting a new center value of a polygon you should move all vertices accordignly, that is missing here.

src_c/polygon.c
```diff
@@ -358,7 +467,68 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+static int
+pg_polygon_set_center(pgPolygonObject *self, PyObject *value, void *closure)
+{
+    DEL_ATTR_NOT_SUPPORTED_CHECK_NO_NAME(value);
+    if (!pg_TwoDoublesFromObj(value, &(self->polygon.c_x), &(self->polygon.c_y))) {
+        PyErr_SetString(PyExc_TypeError, "Expected a sequence of 2 numbers");
+        return -1;
+    }
+    return 0;
+}
```
itzpr3d4t0r, line 523: You are changing the center values but not the vertices positions, you should also change the vertices positions.

itzpr3d4t0r (changes requested):

src_c/polygon.c
itzpr3d4t0r: Basically instead of running through a for loop for N vertices and then running through another N vertices just to get the center, you could do everything inside the first loop. This is not feasable with the memcpy strategies

src_c/polygon.c
maqa41: Done! 

FYI: I will write "Done!" for the ones I finish and then commit changes.

src_c/polygon.c
maqa41: Done!

src_c/polygon.c
maqa41: Done!
Also added an internal function called "move_polygon" to help me out in moving polygon in all of our functions where we need to move. This way will make sure we aren't doing anything wrong as in set center functions.

src_c/polygon.c
maqa41: Done!

src_c/polygon.c
maqa41: Done!

src_c/polygon.c
maqa41: Done!

src_c/polygon.c
maqa41: Done!

src_c/polygon.c
maqa41: Done!

src_c/polygon.c
maqa41: Done!

.gitignore
```diff
@@ -129,3 +129,14 @@ dmypy.json
...
+.vs/slnx.sqlite
+.vs/ProjectSettings.json
+.vs/newpygamestuff/v17/ipch/AutoPCH/3cc4178b25cf09d/GEOMETRY.ipch
+.vs/newpygamestuff/v17/Browse.VC.opendb
+.vs/newpygamestuff/v17/Browse.VC.db-wal
+*.db-shm
+.vs/newpygamestuff/v17/Browse.VC.db
+.vs/newpygamestuff/v17/.suo
+.vs/newpygamestuff/FileContentIndex/read.lock
+*.vsidx
```
Emc2356, line 142: you can just gitignore the whole content of `.vs/` folder

src_c/geometry.c
```diff
@@ -121,3 +121,5 @@ MODINIT_DEFINE(geometry)
     }
     return module;
 }
+
+
```
Emc2356, line 125: needs formating, always format code before commiting

src_c/polygon.c
```diff
@@ -7,6 +7,24 @@
 #include <stddef.h>
 #include <math.h>
 
+static int
+set_polygon_center_coords(pgPolygonBase *polygon)
```
Emc2356, line 11: i would prefer if this was named `_set_polygon_center_coords` to indicate that is private

src_c/polygon.c
```diff
@@ -94,7 +112,8 @@ pgPolygon_FromObject(PyObject *obj, pgPolygonBase *out)
 
         memcpy(out->vertices, poly->vertices,
                poly->verts_num * 2 * sizeof(double));
-
+        /* Function to set polygon center*/
```
Emc2356, line 115: already explained by the name of the functions, dont add comments if the function name already covers it

src_c/polygon.c
```diff
@@ -156,7 +177,8 @@ pgPolygon_FromObject(PyObject *obj, pgPolygonBase *out)
                 Py_DECREF(tmp);
                 tmp = NULL;
             }
-
+            /*Function to set polygon center*/
```
Emc2356, line 180: again, uneeded comment

src_c/polygon.c
```diff
@@ -271,6 +293,8 @@ _pg_polygon_subtype_new2(PyTypeObject *type, double *vertices,
                verts_num * 2 * sizeof(double));
 
         polygon_obj->polygon.verts_num = verts_num;
+        /*Function to set polygon center*/
```
Emc2356, line 296: and here

src_c/polygon.c
```diff
@@ -271,6 +293,8 @@ _pg_polygon_subtype_new2(PyTypeObject *type, double *vertices,
...
+        set_polygon_center_coords(&(polygon_obj->polygon));
```
Emc2356, line 297: uneeded parens, in C `->` operator has priority over the addres-of operator

src_c/polygon.c
```diff
@@ -341,7 +367,90 @@ pg_polygon_copy(pgPolygonObject *self, PyObject *_null)
     return pgPolygon_New(&self->polygon);
 }
 
+static PyObject *
+pg_polygon_move(pgPolygonObject *self, PyObject *const *args, Py_ssize_t nargs)
+{
+    double x = 0, y = 0;
+    switch (nargs) {
+        case 1:
+            if (PySequence_Check(args[0])) {
```
Emc2356, line 376: you can directly call `pg_TwoDoublesFromObj` without the extra sequence check

src_c/polygon.c
```diff
@@ -341,7 +367,90 @@ pg_polygon_copy(pgPolygonObject *self, PyObject *_null)
     return pgPolygon_New(&self->polygon);
 }
 
+static PyObject *
+pg_polygon_move(pgPolygonObject *self, PyObject *const *args, Py_ssize_t nargs)
+{
+    double x = 0, y = 0;
```
Emc2356, line 373: i would say call then Dx, Dy and not just x, y
as i see it with x, y you will be setting the coordinates but with Dx, Dy you are calculating some difference

maqa41: I already changed whole function so no worries there.

src_c/polygon.c
maqa41: I think this one also is from move function, so I have changed that one as itzpr said me to do

src_c/polygon.c
itzpr3d4t0r: Agreed

itzpr3d4t0r:
Closing this as it's now a duplicate

PR #88, opened 2022-09-22
Added midpoint, midpoint_x, midpoint_y attributes to line.

itzpr3d4t0r:


src_c/line.c
```diff
@@ -234,7 +234,8 @@ pg_line_copy(pgLineObject *self, PyObject *_null)
 }
 
 static PyObject *
-pg_line_is_parallel(pgLineObject *self, PyObject *const *args, Py_ssize_t nargs)
+pg_line_is_parallel(pgLineObject *self, PyObject *const *args,
```
novialriptide, line 237: Was this changed via a linter? This change should be reverted.

test/test_line.py
```diff
@@ -285,6 +285,92 @@ def test_attrib_b(self):
...
+
+        self.assertEqual(line.midpoint_x, expected_midpoint[0])
+
+        line.midpoint_x = expected_midpoint[0] - 1
+
+        self.assertEqual(line.midpoint_x, expected_midpoint[0] - 1)
+
+        line = Line(0, 0, 1, 0)
+
+        for value in (None, [], "1", (1,), [1, 2, 3]):
```
novialriptide, line 339: We should create a standardized variable for checking these integers & floats in the future, this seems to be a common copy-paste scenario.

novialriptide (changes requested):
Just one change

itzpr3d4t0r:
I don't know what i did here i admit nothing was intentional

src_c/line.c
itzpr3d4t0r: Yes this was my IDE automatic clang formatter

test/test_line.py
itzpr3d4t0r: YEah i agree, some tests are a little more specific tho, so we'll most likely need more than one variant

Issue #89, opened 2022-09-23
Polygon.update_vertex()


ScriptLineStudios:
I think a function to update one of the vertices of a Polygon would be useful. I propose something like this.

```python
polygon = geometry.Polygon([(100, 10), (600, 600), (10, 600)])
polygon.update_vertex(0, (200, 10))
```

itzpr3d4t0r:
I don't know if a function is the correct way of handling this, it may be sufficient to make the polygon subscriptable like we have for line:
```Python
polygon = geometry.Polygon([(100, 10), (600, 600), (10, 600)])
polygon[0] = (200, 10)
```

ScriptLineStudios:
`

> I don't know if a function is the correct way of handling this, it may be sufficient to make the polygon subscriptable like we have for line:
> 
> ```python
> polygon = geometry.Polygon([(100, 10), (600, 600), (10, 600)])
> polygon[0] = (200, 10)
> ```

Actually yeah that makes a lot more sense.

ScriptLineStudios:
I will try to do this

PR #90, opened 2022-09-23
Subscriptable Polygons

ScriptLineStudios:
Subscriptable polygons to allow for updating vertices based on @itzpr3d4t0r suggestion in #89 Also added a new example and appropriate test.


Emc2356 (changes requested):
if we will be going down this path then we might as well implement the full sequence protocol,
for reference you can check out the Line class in line.c

src_c/polygon.c
```diff
@@ -305,6 +305,129 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+    if (i < 0) {
+        i += poly->verts_num;
+    }
+
+    if (i < 0 || i > poly->verts_num) {
+        PyErr_SetString(PyExc_IndexError, "Invalid vertex Index");
+        return -1;
+    }
+
+    double v1, v2;
```
Emc2356, line 322: i would prefer if this where called Vx, Vy

src_c/polygon.c
```diff
@@ -305,6 +305,129 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+    poly->vertices[i * 2] = v1;
+    poly->vertices[i * 2 + 1] = v2;
+    return 0;
+}
+
+static Py_ssize_t
+pg_polygon_seq_length(pgPolygonObject *self)
+{
+    pgPolygonBase *poly = &self->polygon;
+    return poly->verts_num;
```
Emc2356, line 336: use `pgPolygon_GETVERTSNUM`
```suggestion
    return pgPolygon_GETVERTSNUM(self);
```

src_c/polygon.c
```diff
@@ -305,6 +305,129 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+static int
+pg_polygon_contains_seq(pgPolygonObject *self, PyObject *arg)
+{
+    pgPolygonBase *poly = &self->polygon;
+
+    if (PyTuple_Check(arg) || PyList_Check(arg)) {
+        arg = PySequence_Tuple(arg);
+        Py_ssize_t i;
+        for (i = 0; i < poly->verts_num; i++) {
+            PyObject *tup = pg_TupleFromDoublePair(
```
Emc2356, line 395: no reason to pack the verticies into a tuple and then unpacking it
also if this would make it in this is a memory leak as you are allocating the PyObject* and not deleting it

Emc2356 (commented):
i found this problems with a sweap, just handle this ones and then i will recheck it

src_c/polygon.c
ScriptLineStudios: The reason I packed arg into a tuple first is so that the function can also accept lists. And then I can just use the TupleFromDoublePair function. Would it be okay to keep this if I deallocate PyObject* once I have finished?

src_c/polygon.c
```diff
@@ -305,6 +305,131 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+    else {
+        PyErr_SetString(PyExc_TypeError, "Expected a number or sequence");
+        return -1;
+    }
+}
+
+static PyObject *
+pg_polygon_subscript(pgPolygonObject *self, PyObject *op)
+{
+    PyObject *vertices = _pg_polygon_vertices_aslist(&self->polygon);
```
itzpr3d4t0r, line 363: `_pg_polygon_vertices_aslist` can fail and returns `NULL` in that case. you should check if that's NULL and just return NULL.
```suggestion
    PyObject *vertices = _pg_polygon_vertices_aslist(&self->polygon);
    if (!vertices) {
        return NULL;
    }
```

src_c/polygon.c
```diff
@@ -305,6 +305,131 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+    if (i < 0) {
+        i += poly->verts_num;
+    }
+
+    if (i < 0 || i > poly->verts_num) {
+        return RAISE(PyExc_IndexError, "Invalid vertex Index");
+    }
+
+    return PyList_GetItem(vertices, i);
+}
```
itzpr3d4t0r, line 384: I'm no expert on these but i suspect this is wasting a lot of performance? Why creating a whole list with N tuples and 2N PyFloats to only extract one?
Isn't it better to just evaluate the passed index and create a single tuple with the right values?

src_c/polygon.c
```diff
@@ -305,6 +305,131 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+
+static int
+pg_polygon_contains_seq(pgPolygonObject *self, PyObject *arg)
+{
+    pgPolygonBase *poly = &self->polygon;
+
+    if (PyTuple_Check(arg) || PyList_Check(arg)) {
```
itzpr3d4t0r, line 391: Use `PySequence_FAST_CHECK` here
```suggestion
    if (PySequence_FAST_CHECK (arg)) {
```

Emc2356: good catch

src_c/polygon.c
```diff
@@ -305,6 +305,131 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+                return 1;
+            }
+        }
+        return 0;
+    }
+    else {
+        PyErr_SetString(PyExc_TypeError, "Expected a tuple");
+        return -1;
+    }
+}
```
itzpr3d4t0r, line 419: I don't like this either, as @Emc2356 suggested, this is unnecessary complexity, you can convert the input x and y values once and then compare them with any other double contained in the vertices:
See if this works

```C
static int
pg_polygon_contains_seq(pgPolygonObject *self, PyObject *arg)
{
    if (!PySequence_FAST_CHECK(arg)) {
        PyErr_SetString(PyExc_TypeError, "Expected a tuple");
        return -1;
    }

    pgPolygonBase *poly = &self->polygon;

    double x, y;
    if (!pg_TwoDoublesFromObj(arg, &x, &y)) {
        PyErr_SetString(PyExc_TypeError, "Expected a sequence of 2 numbers");
        return -1;
    }

    Py_ssize_t i;
    for (i = 0; i < poly->verts_num; i += 2) {
        if (poly->verts[i] == x && poly->verts[i + 1] == y) {
            return 1;
        }
    }

    return 0;
}
```

Emc2356: i say we use the one that i sent on discord, it is better as it covers the sequence protocol and not just tuple and list

src_c/polygon.c
```diff
@@ -305,6 +305,150 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+    }
+
+    return pg_TupleFromDoublePair(poly->vertices[i * 2], poly->vertices[i * 2 + 1]);
+}
+
+static int
+pg_polygon_contains_seq(pgPolygonObject *self, PyObject *arg)
+{
+    double x, y;
+    if (PySequence_FAST_CHECK(arg)) {
```
itzpr3d4t0r, line 388: Invert this condition and put the error set and return inside. In this way we reduce the indentation level by one.

src_c/polygon.c
```diff
@@ -437,6 +437,148 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+        }
+        Py_DECREF(tmp);
+
+        pgPolygonBase *poly = &self->polygon;
+        Py_ssize_t i;
+        for (i = 0; i < poly->verts_num; i+=2) {
+            if (poly->vertices[i] == x && poly->vertices[i+1] == y) {
+                return 1;
+            }
+        }
```
itzpr3d4t0r, line 561: This should be outside of all if statements, not iside the second if, in this manner it will only work if you pass it any sequence apart from lists and tuples.

src_c/polygon.c
```diff
@@ -437,6 +437,148 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+    }
+    else {
+        PyErr_SetString(PyExc_TypeError, "Expected a sequence");
+        return -1;
+    }
+    return 0;
+}
```
itzpr3d4t0r, line 568: This function can be massively simplified when #100 will be merged.

src_c/polygon.c
itzpr3d4t0r: And it's flawed already, at first you're discarding sequences that are not tuples or lists, but later you check for sequences.

src_c/polygon.c
itzpr3d4t0r: And it's flawed already, at first you're discarding sequences that are not tuples or lists, but later you check for sequences.

itzpr3d4t0r: When #100 is merged we could have something as simple as this:
```C
static int
pg_polygon_contains_seq(pgPolygonObject *self, PyObject *arg)
{
    double x, y;
    
    if (!pg_TwoDoublesFromObj(arg, &x, &y)) {
        PyErr_SetString(PyExc_TypeError, "Expected a sequence of 2 numbers");
        return -1;
    }
    
    pgPolygonBase *poly = &self->polygon;
    Py_ssize_t i;
    for (i = 0; i < poly->verts_num; i+=2) {
        if (poly->vertices[i] == x && poly->vertices[i+1] == y) {
            return 1;
        }
    }
    
    return 0;
}
```

src_c/polygon.c
```diff
@@ -437,6 +437,115 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+    }
+}
+
+static PyObject *
+pg_polygon_subscript(pgPolygonObject *self, PyObject *op)
+{
+    PyObject *index = PyNumber_Index(op);
+    Py_ssize_t i;
+
+    if (index == NULL) {
```
itzpr3d4t0r, line 497: I'd say use `!index` here

test/test_polygon.py
```diff
@@ -91,6 +91,43 @@ def test_construction_frompolygon(self):
...
 
+    def test_subscript(self):
+        """Checks whether reassigning a vertex works correctly"""
+        po = Polygon([p1, p2, p3, p4])
+
+        po[0] = (70.0, 80.0)
+        self.assertEqual(po[0], (70.0, 80.0))
+
+        po[1] = [45.0, 38.0]
+        self.assertEqual(po[1], (45.0, 38.0))
```
itzpr3d4t0r, line 102: I'd like to see:
- [ ] Every valid index tested, both positive and negative ( `0->3` and `-1 -> -3` )
- [ ] Invalid positive index tested, both super close to the list size and super far
- [ ] Invalid negative index ( in this case i.e. -4 and -100)
- [ ] Assigning a vertex to itself won't change the vertex

test/test_polygon.py
```diff
@@ -91,6 +91,43 @@ def test_construction_frompolygon(self):
...
+
+    def test_length(self):
+        po = Polygon([p1, p2, p3, p4])
+        self.assertEqual(len(po), 4)
+
+        po = Polygon([p1, p2, p3, p4, [75.0, 83.0], [23.0, 12.0], [90.0, 134.0]])
+        self.assertEqual(len(po), 7)
```
itzpr3d4t0r, line 109: We have many ways to construct a polygon, might be wise to test all of them to check the size is the same always.

test/test_polygon.py
```diff
@@ -91,6 +91,43 @@ def test_construction_frompolygon(self):
...
+
+    def test_contains(self):
+        po = Polygon([p1, p2, p3, p4])
+        self.assertTrue(p1 in po)
+        self.assertTrue(p2 in po)
```
itzpr3d4t0r, line 114: Again, test for all vertices, not just two

test/test_polygon.py
```diff
@@ -91,6 +91,43 @@ def test_construction_frompolygon(self):
...
+
+        self.assertFalse([90.0, 47.0] in po)
+        self.assertFalse((35.0, 9.0) in po)
+
+        invalid_types = (None, [], "1", (1,), Vector3(1, 1, 1), 1)
```
itzpr3d4t0r, line 119: Add a `(1, 2, 3)` tuple to these

test/test_polygon.py
```diff
@@ -91,6 +91,43 @@ def test_construction_frompolygon(self):
...
+
+        for value in invalid_types:
+            with self.assertRaises(TypeError):
+                value in po
+
+        invalid_values = ([17.0, None], ["1", 47.0], (None, None), ("123", "456"))
```
itzpr3d4t0r, line 125: Add (17.0, "12")

itzpr3d4t0r (approved):
LGTM, just a minor thing

PR #92, opened 2022-09-23
Added center attribute & move()/move_ip()

maqa41:
here we have everything working on. Added tests are also working.


src_c/polygon.c
```diff
@@ -7,6 +7,40 @@
...
+        sum_x += polygon->vertices[i2];
+        sum_y += polygon->vertices[i2 + 1];
+    }
+    polygon->c_x = sum_x / polygon->verts_num;
+    polygon->c_y = sum_y / polygon->verts_num;
+    return 1;
+}
+
+static int
+move_polygon(pgPolygonBase *polygon, double dx, double dy)
```
Emc2356, line 29: this should be `_pg_move_polygon_helper`

src_c/polygon.c
```diff
@@ -140,23 +187,31 @@ pgPolygon_FromObject(PyObject *obj, pgPolygonBase *out)
 
         if (length >= 3) {
             Py_ssize_t i;
+            Py_ssize_t i2;
```
Emc2356, line 190: i am not sure about that but the `s` in ssize is for signed, you need to use unsigned int64 because memory addresses are unsigned

src_c/polygon.c
```diff
@@ -212,6 +267,7 @@ pgPolygon_FromObjectFastcall(PyObject *const *args, Py_ssize_t nargs,
     }
     else if (nargs >= 3) {
         Py_ssize_t i;
+        Py_ssize_t i2;
```
Emc2356, line 270: again here

Emc2356:
also needs formating

src_c/polygon.c
```diff
@@ -341,7 +405,64 @@ pg_polygon_copy(pgPolygonObject *self, PyObject *_null)
...
+            goto error;
+        }
+    }
+    else {
+        goto error;
+    }
+
+    move_polygon(&self->polygon, Dx, Dy);
+    return _pg_polygon_subtype_new2(Py_TYPE(self), self->polygon.vertices,
+                                    self->polygon.verts_num);
```
itzpr3d4t0r, line 430: This move function is wrong, you don't move the original polygon, clone it and return it.
What you instead do is (in this case) create a heap allocated double * memory of the correct size, copy the original polygon vertices over, move them and then create the new polygon with that memory. At last you free the original memory.

itzpr3d4t0r, line 430:
```suggestion
    double *verts = PyMem_New(double, self->polygon.verts_num*2);
    if (!verts) {
        return NULL;
    }
    memcpy(verts, self->polygon.vertices, self->polygon.verts_num * 2 * sizeof(double))
    
    // move the vertices inside the double *verts array here
    
    PyObject *tmp = _pg_polygon_subtype_new2(Py_TYPE(self), verts, self->polygon.verts_num);
    if (!tmp) {
        PyMem_Free(verts);
        return NULL;
    }
    PyMem_Free(verts);
    return tmp;
```

src_c/polygon.c
maqa41: hmmm... what you are saying is that since that `i2` variable is not used here in particular for python function implementation, I should use size_t, right?

src_c/polygon.c
maqa41: Done!

src_c/polygon.c
maqa41: Done!

src_c/polygon.c
```diff
@@ -305,7 +437,71 @@ pg_polygon_get_vertices(pgPolygonObject *self, void *closure)
...
+        return -1;
+    }
+    _pg_move_polygon_helper(&(self->polygon), 0.0, (val - self->polygon.c_y));
+    return 0;
+}
+
+static PyObject *
+pg_polygon_get_center(pgPolygonObject *self, void *closure)
+{
+    return pg_tuple_from_values_double(self->polygon.c_x, self->polygon.c_y);
```
itzpr3d4t0r, line 481: You should use the new `pg_TupleFromDoublePair` here

test/test_polygon.py
```diff
@@ -124,20 +125,205 @@ def test_copy_invalid_args(self):
...
+
+        vertices = [tuple(vertex) for vertex in vertices]
+        self.assertEqual(poly.vertices, vertices)
+        self.assertEqual(poly.c_x, 200.0)
+        self.assertEqual(poly.c_y, 200.0)
+
+    def test_center__invalid_value(self):
+        """Ensures the function can handle the polygon center component by invalid data types."""
+        poly = Polygon(_some_vertices.copy())
+        for value in (None, [], "1", (1,), [1, 2, 3]):
```
itzpr3d4t0r, line 210: Might be wise to test for a valid type (a tuple of two elements, but one of the elements is wrong, like:
(1, "s"), (None, 3), (2, (3))

maqa41: wow, this makes really sense. Alright let me patch this up.

Emc2356 (approved):
LGMT!

test/test_polygon.py
```diff
@@ -124,20 +125,213 @@ def test_copy_invalid_args(self):
...
+
+        for vertex in vertices:
+            vertex[0] += 10.0
+            vertex[1] += 10.0
+
+        vertices = [tuple(vertex) for vertex in vertices]
+        self.assertEqual(vertices, new_poly.vertices)
+        self.assertEqual(new_poly.c_x, center_x + 10.0)
+        self.assertEqual(new_poly.c_y, center_y + 10.0)
+
```
itzpr3d4t0r, line 237: I'd like one more test that ensures that using polygon.move on a polygon won't move the original polygon

test/test_polygon.py
```diff
@@ -124,20 +125,213 @@ def test_copy_invalid_args(self):
...
+        vertices = _some_vertices.copy()
+        for vertex in vertices:
+            vertex[0] += 200.0 - poly_center[0]
+            vertex[1] += 200.0 - poly_center[1]
+
+        vertices = [tuple(vertex) for vertex in vertices]
+        self.assertEqual(poly.vertices, vertices)
+        self.assertEqual(poly.c_x, 200.0)
+        self.assertEqual(poly.c_y, 200.0)
+
```
itzpr3d4t0r, line 206: I'd like a test that ensures that setting the polygon center to the current polygon center won't move any of the vertices

PR #97, opened 2022-09-24
Disallow degenerate line creation

### LINKED ISSUE #96: Disallow creating a degenerate line segment
Right now we allow having a "point" line segment:
```Python
line = Line(1, 1, 1, 1)
```
This has to be fixed. many times we divide by `x2-x1` or `y2-y1` and that will error out.
My idea is to implement a new check function to add inside the pgLine_FromObject function after the line attributes are set that checks their validity. if not it will error out.

itzpr3d4t0r:
Resolves #96.

### DIFF
--- a/src_c/line.c
+++ b/src_c/line.c
@@ -7,6 +7,8 @@
 #include <stddef.h>
 #include <math.h>
 
+#define IS_LINE_VALID(line) (line->x1 != line->x2 || line->y1 != line->y2)
+
 static inline double
 pgLine_Length(pgLineBase *line)
 {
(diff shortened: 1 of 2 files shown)


src_c/line.c
```diff
@@ -765,8 +770,7 @@ static PyMappingMethods pg_line_as_mapping = {
 static int
 pg_line_bool(pgLineObject *self)
 {
-    return self->line.x1 != 0 || self->line.y1 != 0 || self->line.x2 != 0 ||
-           self->line.y2 != 0;
+    return 1;
```
itzpr3d4t0r, line 773: This had to be changed because no line can have all zeroes as positions now.

PR #104, opened 2022-09-25
implemented line.midpoint, line.midpoint_x, line.midpoint_y attributes

itzpr3d4t0r:
reup of #88 that got obliterated by accident.


Emc2356 (commented):
i like this PR but i dont think we should be able to set the midpoints
i would prefer if they were readonly attributes

itzpr3d4t0r:
> i like this PR but i dont think we should be able to set the midpoints i would prefer if they were readonly attributes

Well what if when you create a line you want that line to be positioned center to a point?
You'd have to invent some weir move functions that moves the line to be centered. with this you just do `line.midpoint = (3, 3)` and it will all be automatically handled. I don't see confusing scenarios here...

Emc2356:
> > i like this PR but i dont think we should be able to set the midpoints i would prefer if they were readonly attributes
> 
> Well what if when you create a line you want that line to be positioned center to a point? You'd have to invent some weird move function that moves the line to be centered. with this you just do `line.midpoint = (3, 3)` and it will all be automatically handled. I don't see confusing scenarios here...

well you can do something like this: `line.move(-line.midpoint_x, line.midpoint_y)`, it is simple and i dont like the idea of the setters

PR #105, opened 2022-09-25
Line.angle attribute 

ScriptLineStudios:
An angle attribute for Lines


src_c/line.c
```diff
@@ -7,6 +7,9 @@
 #include <stddef.h>
 #include <math.h>
 
+#define PI 3.14159265358979323846264
```
Emc2356, line 10: wrap it in #ifndef PI as PI my be defined in other places also

src_c/line.c
```diff
@@ -7,6 +7,9 @@
...
+#define RAD_TO_DEG(x) (x * 180 / PI)
```
Emc2356, line 11: again, add this in #ifndef

src_c/line.c
```diff
@@ -916,11 +919,27 @@ pg_line_setb(pgLineObject *self, PyObject *value, void *closure)
...
+{
+    double dx = self->line.x2 - self->line.x1;
+    double dy = self->line.y2 - self->line.y1;
+
+    if (dx == 0.0) {
+        return (self->line.y2 > self->line.y1) ? PyFloat_FromDouble(-90.0)
+                                               : PyFloat_FromDouble(90.0);
+    }
+
+    double gradient = (dy / dx);
```
Emc2356, line 933: why parens

Emc2356 (commented):
just this things and i think this is fine

src_c/line.c
```diff
@@ -949,11 +956,26 @@ pg_line_setb(pgLineObject *self, PyObject *value, void *closure)
     return -1;
 }
 
+static PyObject *
+pg_line_getangle(pgLineObject *self, void *closure)
+{
+    double dx = self->line.x2 - self->line.x1;
+    double dy = self->line.y2 - self->line.y1;
```
itzpr3d4t0r, line 963: Not that it matters much, but you should move this `dy` calclation after the if statement.

PR #110, opened 2022-09-27
Polygon.perimeter

ScriptLineStudios:
As per pygame.geometry project board


src_c/polygon.c
```diff
@@ -590,6 +590,25 @@ pg_polygon_get_center(pgPolygonObject *self, void *closure)
...
+    double perimeter = 0;
+    for (int i = 0; i < poly->verts_num - 1; i++) {
+        double Vx = poly->vertices[i * 2];
+        double Vy = poly->vertices[i * 2 + 1];
+
+        double Vx2 = poly->vertices[(i + 1) * 2];
+        double Vy2 = poly->vertices[(i + 1) * 2 + 1];
+
+        perimeter += sqrt((Vx2 - Vx) * (Vx2 - Vx) + (Vy2 - Vy) * (Vy2 - Vy));
+    }
```
itzpr3d4t0r, line 607: This is incorrect because you're not adding the last edge. first i would add a quick distance function like this ( this would simplify the inner loop a bit
```C
static inline double
_pg_distance(double x1, double y1, double x2, double y2)
{
    double dx = x2 - x1;
    double dy = y2 - y1;
    return sqrt(dx * dx + dy * dy);
}
```
Then i would do something like this
```C
    double *vertices = self->polygon.vertices;
    Py_ssize_t i;
    for (i = 0; i < poly->verts_num - 1; i++) {
        perimeter +=
            _pg_distance(vertices[i * 2], vertices[i * 2 + 1],
                         vertices[(i + 1) * 2], vertices[(i + 1) * 2 + 1]);
    }

    perimeter += _pg_distance(vertices[i * 2], vertices[i * 2 + 1],
                              vertices[0], vertices[1]);
```

Emc2356: should use `PG_FORCEINLINE` or `SDL_FORCE_INLINE`

PR #114, opened 2022-09-28
New polygon creation method

itzpr3d4t0r:
This PR adds some more polygon benchmarks and a new function that improves performance for the polygon move functions and for any future function that returns a copy of the original polygon with updated vertices.
This is because of the fact that in those kinds of functions we allocate new memory that we use for vertices manipulation already, so it doesn't make sense to create new memory again and discard the other. All of this saves an extra memory allocation and a memcpy, saving about 20-25%.


novialriptide (approved):
LGTM. I'm still not set on `regular_polygon()`, but this is good for now.

PR #123, opened 2022-10-02
Disallowed degenerate(radius=0) circles + better error messages

itzpr3d4t0r:
Before this PR you could do:

```Python
circle = Circle((10, 10), 4)
circle.r = 0
circle.d = 0
circle.circumference = 0
circle.area = 0
```
This is no more.

Another issue this PR tries to solve is attribute assignment error messages. These were very imprecise and did not distinguish between a wrong type being passed and  a wrong value. It now raises `TypeError` on wrong type and `ValueError` on wrong value ( <= 0 ).


novialriptide (approved):
LGTM

Issue #58, opened 2022-08-29
Move `Line.raycast()` to `raycast()`


novialriptide:
Something like this? This is a low priority for now, but I highly suggest this issue must be worked on before we migrate to https://github.com/pygame/pygame since `Ray != Line`
As @Starbuck5 suggested, we can do a single function that's not attached to any instance such as
```python
raycast(
    start_pos: Sequence[int, int],
    endpoint: Union[Sequence[float, float], None] = None,
    angle: Union[float, None] = None
)
# If the angle and endpoint both have values or are None, it'll raise an error.
```

The re-implementation should be pretty simple since the C implementation already exists.

Starbuck5:
It seems to me that raycast should just be an independent function in the geometry module.

novialriptide:
How should we design this function though? Should we take in `end_point: Sequence[float, float]`, `angle: float`, or `direction: Sequence[float, float]`?

novialriptide:
I just got started on this issue. Expect a PR in a couple days.

PR #126, opened 2022-10-06
Line.raycast() to raycast() relocation

novialriptide:
There will be another follow-up PR that will fix https://github.com/novialriptide/pygame_geometry/issues/127 after this has been merged.

Fixes: https://github.com/novialriptide/pygame_geometry/issues/58

This PR adds a new feature that supports angles for raycasting as well.


src_c/geometry.c
```diff
@@ -7,6 +7,108 @@
...
+    double angle;
+    double max_dist;
+    double target_pos[2] = {0, 0};
+
+    static char *keywords[] = {"start_pos", "collisions", "endpoint", "angle", "max", NULL};
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "OO|Odd",
+        keywords, &startpoint, &collisions, &endpoint, &angle, &max_dist))
+    {
+        return NULL;
+    }
```
itzpr3d4t0r, line 27: As i said in discord i'm not completely fine with kwargs here. There should be more thought put into this. 
### 1
You have two parameters for start  and end positions, where one is way at the start and the other way at the back. We could be smarter than this and have the function accept as first argument either a sequence of `(start_pos, end_pos)` (meaning we have a tuple containing two tuples) or a single value (tuple) containing the ray origin. The function should be resposible to check the type that's passed and act accordingly. In this way we reduce the param. number by one.

### 2
The `start_pos` and `collisions` parameters should not be kwargs, as you always need at least these two, and they should always be at the start of the parameter list passed for grammatical coherence and clarity. So you should remove them from the `static char *keywords[]` array.

### 3
We could further remove parameters if we consider that a ray can be represented in two ways (at least for what we have right now) which are:
1. A Sequence containing two tuples which represent the start and end position of the ray. It's obviout that in this case the ray already has a max_length that's iplicit to the Line length. `((x1, y1), (x2, y2))`
2. We could also imagine a ray as a **Sequence of start_pos, angle, max_distance**. In that case we would simply represent the ray as a Tuple of `(start_pos, angle, max_distance)` or `((x, y), angle, max_distance)`

Given these two options it's apparent how we wouldn't need to pass the extra `endpoint`,  `angle`, `max` parameters as we could simply have a single parameter that represents both: a `ray` parameter if you wish, that's automatically calculated and converted to a usable object.

So combining all that i said we would have something like this:

```C
    static char *keywords[] = {"ray", "collisions", NULL};
    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "OO", keywords, &ray, &collisions))
    {
        return NULL;
    }
```
As you can see now we can't have kwargs anymore as it would be superflous. At this point you could revert it back to FASTCALL and have the function take care of the first argument.

src_c/geometry.c
```diff
@@ -7,6 +7,108 @@
...
+
+    farr = PySequence_Fast_ITEMS(collisions);
```
itzpr3d4t0r, line 29: Idk why but you removed the Check that makes sure that the object being passed to the function is a PySequence_Fast compatible sequence.

src_c/geometry.c
```diff
@@ -7,6 +7,108 @@
...
+
+    if (!PySequence_Check(startpoint)) {
+        return RAISE(PyExc_TypeError, "positions must be sequences");
+    }
+
+    pgLineObject *line = (pgLineObject *)pgLine_Type.tp_new(&pgLine_Type, NULL, NULL);
```
itzpr3d4t0r, line 35: You don't need to create a PyObject here, just create the underlying `pgLineBase` struct.

src_c/geometry.c
```diff
@@ -7,6 +7,108 @@
...
+    if (line) {
+        line->line.x1 = PyFloat_AsDouble(PySequence_GetItem(startpoint, 0));
+        line->line.y1 = PyFloat_AsDouble(PySequence_GetItem(startpoint, 1));
```
itzpr3d4t0r, line 38: This is wrong, PySequence_GetItem is massively inefficient ( we have functions for extracting doubles from PyObject *) and  PyFloat_AsDouble can fail, i see no check here... 
You should use our base functions for this.

src_c/geometry.c
```diff
@@ -7,6 +7,108 @@
...
+    }
+    
+    Py_ssize_t arg0_length = PySequence_Size(startpoint);
+    Py_ssize_t col_length = PySequence_Size(collisions);
```
itzpr3d4t0r, line 42: Once you know that `collisions` is a PySequence_Fast object you can call the much more efficient `PySequence_Fast_GET_SIZE` macro.

src_c/geometry.c
```diff
@@ -7,6 +7,108 @@
...
+    if (endpoint != NULL) {
+        if (!PySequence_Check(endpoint)) {
+            return RAISE(PyExc_TypeError, "positions must be sequences");
+        }
+        Py_ssize_t arg1_length = PySequence_Size(endpoint);
+        if (arg1_length != 2) {
+            return RAISE(PyExc_TypeError, "incorrect end position size");
+        }
+        line->line.x2 = PyFloat_AsDouble(PySequence_GetItem(endpoint, 0));
+        line->line.y2 = PyFloat_AsDouble(PySequence_GetItem(endpoint, 1));
```
itzpr3d4t0r, line 57: Again this has the same problem as before

src_c/geometry.c
```diff
@@ -7,6 +7,108 @@
...
+    }
+    else {
+        double offset1 = cos(angle * PI / 180) * max_dist;
+        double offset2 = sin(angle * PI / 180) * max_dist;
+        line->line.x2 = line->line.x1 - offset1;
+        line->line.y2 = line->line.y1 - offset2;
```
itzpr3d4t0r, line 63: this line->line this will not be needed once you create a pgLineBase struct instead of a PyObject

geometry.pyi
```diff
@@ -194,3 +191,16 @@ class Polygon:
...
+    collisions: Sequence[Circle, Line],
+    endpoint: Union[Coordinate, None] = None,
+) -> Union[Coordinate, None]: ...
+@overload
+def raycast(
+    startpoint: Coordinate,
+    collisions: Sequence[Circle, Line],
+    angle: Union[float, None] = None,
+    max: Union[float, None] = None,
+) -> Union[Coordinate, None]: ...
```
itzpr3d4t0r, line 206: All this needs to be changed as a result

src_c/geometry.c
```diff
@@ -74,6 +176,7 @@ geometry_regular_polygon(PyObject *_null, PyObject *const *args,
 static PyMethodDef _pg_module_methods[] = {
     {"regular_polygon", (PyCFunction)geometry_regular_polygon, METH_FASTCALL,
      NULL},
+    {"raycast", (PyCFunction)pg_raycast, METH_VARARGS | METH_KEYWORDS, NULL},
```
itzpr3d4t0r, line 179: METH_FASTCALL here

src_c/geometry.c
```diff
@@ -7,6 +7,108 @@
...
+pg_raycast(PyObject *self, PyObject *args, PyObject *kwargs)
+{
+    PyObject *startpoint;
+    PyObject *endpoint = NULL;
+    PyObject **farr;
+    PyObject **collisions;
+    Py_ssize_t loop;
+    double angle;
+    double max_dist;
+    double target_pos[2] = {0, 0};
```
itzpr3d4t0r, line 20: you don't need to specify the array size if you already provide the array contents like that.

src_c/geometry.c
novialriptide: I don't like the idea of having tuples inside tuples, would this be a better API design instead? This will use FASTCALL instead.
```python
@overload
def raycast(originpos, endpos, colliders): ...
@overload
def raycast(originpos, angle, max_dist, colliders): ...
```

src_c/geometry.c
novialriptide: I tried this before but it kept segfaulting earlier, if I remove the args and kwargs parser it might work. Idk.

src_c/geometry.c
```diff
@@ -8,6 +8,108 @@
...
+        farr = PySequence_Fast_ITEMS(args[3]);
+        col_length = PySequence_Fast_GET_SIZE(args[3]);
+    }
+    else {
+        return RAISE(PyExc_TypeError,
+                    "invalid number of arguments");
+    }
+
+
+    pgLineObject *line = (pgLineObject *)pgLine_Type.tp_new(&pgLine_Type, NULL, NULL);
```
itzpr3d4t0r, line 58: You don-t need to create a whole python object, just create a pgLineBase struct here
```suggestion
    pgLineBase line;
```

itzpr3d4t0r: And from here on you should remove all `line->` (s)

src_c/geometry.c
```diff
@@ -8,6 +8,108 @@
...
+    else if (nargs == 4) {
+        if (!pg_TwoDoublesFromObj(args[0], &start_x, &start_y)) {
+            return RAISE(PyExc_TypeError, "the starting position requires a pair of floats");
+        }
+        if (!PyFloat_Check(args[1]) && !pg_DoubleFromObj(args[1], &angle)) {
+            return RAISE(PyExc_TypeError, "invalid angle");
+        }
+        if (!PyFloat_Check(args[2]) && !pg_DoubleFromObj(args[2], &max_dist)) {
+            return RAISE(PyExc_TypeError, "invalid max distance");
+        }
```
itzpr3d4t0r, line 46: Why should these check for PyFloat? There's no need as `pg_DoubleFromObj` already converts floats and ints to double anyway.

src_c/geometry.c
```diff
@@ -8,6 +8,108 @@
...
+        end_x = start_x - cos(angle * PI / 180) * max_dist;
+        end_y = start_y - sin(angle * PI / 180) * max_dist;
+        farr = PySequence_Fast_ITEMS(args[3]);
```
itzpr3d4t0r, line 49: You are not checking that args[3] is a valid `PySequence_Fast` object

src_c/geometry.c
itzpr3d4t0r: Plus there no need to error if someone passes say 10 as angle or distance, we should consider 10 as 10.0 and that's it.

src_c/geometry.c
```diff
@@ -8,6 +8,108 @@
...
+    double start_x, start_y;
+    double end_x, end_y;
+    PyObject **farr;
+    Py_ssize_t col_length;
+    Py_ssize_t loop;
+    double angle;
+    double max_dist;
+
+    if (nargs == 3) {
+        if (!PySequence_Check(args[0]) || !PySequence_Check(args[1]) || !PySequence_FAST_CHECK(args[2])) {
```
itzpr3d4t0r, line 23: I feel these checks are superflous as `pg_TwoDoublesFromObj` ( the statement right after this one ) already checks and errors out if it couldn't retrieve the two values. You are basically doing twice the amount of necessary checks.
```suggestion
        if (!PySequence_FAST_CHECK(args[2])) {
```

novialriptide:
Many changes still needs to be made, I just pushed some changes right now to show some progress.

src_c/geometry.c
```diff
@@ -8,6 +8,113 @@
...
+                                        &temp_t)) {
+                record = MIN(record, temp_t);
+            }
+        }
+        else if (pgRect_Check(farr[loop])) {
+            if (pgIntersection_LineRect(&(line),
+                                        &pgRect_AsRect(farr[loop]), NULL, NULL,
+                                        &temp_t)) {
+                record = MIN(record, temp_t);
+            }
```
itzpr3d4t0r, line 95: Nit. you should remove all the "( )" around the `&(line)` things

novialriptide:
@itzpr3d4t0r Excuse the review re-request, I seem to have forgotten to adjust changes to your reviews. I'll be doing that in a bit.

src_c/geometry.c
itzpr3d4t0r: Remove these, it's not just my feeling, they're actually not needed.

src_c/geometry.c
```diff
@@ -8,6 +8,113 @@
...
+            return RAISE(PyExc_TypeError, "the starting position requires a pair of floats");
+        }
+        if (!pg_TwoDoublesFromObj(args[1], &end_x, &end_y)) {
+            return RAISE(PyExc_TypeError, "the end position requires a pair of floats");
+        }
+        farr = PySequence_Fast_ITEMS(args[2]);
+        col_length = PySequence_Fast_GET_SIZE(args[2]);
+    }
+    else if (nargs == 4) {
+        if (!PySequence_Check(args[0]) || !PySequence_FAST_CHECK(args[3])) {
```
itzpr3d4t0r, line 37: the `PySequence_Check(args[0])` check is not needed as pg_TwoDoublesFromObj errors out already in that case and it has an even better error message.
```suggestion
        if (!PySequence_FAST_CHECK(args[3])) {
```

src_c/geometry.c
```diff
@@ -8,6 +8,113 @@
...
+            return RAISE(PyExc_TypeError, "an internal error has occured");
+        }
+        farr = PySequence_Fast_ITEMS(args[3]);
+        col_length = PySequence_Fast_GET_SIZE(args[3]);
+    }
+    else {
+        return RAISE(PyExc_TypeError,
+                    "invalid number of arguments");
+    }
+
```
itzpr3d4t0r, line 63: Another thing you could do is shorten the code a bit by loading the colliders sequence "once", meaning outside of the ifs. This is feasable because the colliders list is always the last parameter passed, therefore you could just do:

```C
    if (!PySequence_FAST_CHECK(args[nargs-1])) {
            return RAISE(PyExc_TypeError, "Colliders parameter must be a sequence");
        }
    farr = PySequence_Fast_ITEMS(args[nargs-1]);
    col_length = PySequence_Fast_GET_SIZE(args[nargs-1]);
```
with this and the removal of the same parts inside the two ifs we save 5 lines.

novialriptide:
Should we make this function return a tuple instead? The first element will be the actual coordinate value while the second will be a boolean whether or not the ray has actually hit something.

novialriptide:
Here is my proposal on the new returntypes
```python
@overload
def raycast(
    originpos: Coordinate,
    endpos: Coordinate,
    colliders: List[Rect, Circle, Line],
) -> Tuple[Coordinate, boolean]: ...
@overload
def raycast(
    originpos: Coordinate,
    angle: float,
    max_dist: float,
    colliders: List[Rect, Circle, Line],
) -> Tuple[Coordinate, boolean]: ...
```

The `boolean` will be `True` if the ray has hit something, `False` if otherwise.

novialriptide:
@itzpr3d4t0r and I had a discussion over Discord on how we should implement `raycast()`'s collision detector.

I think it would be a good idea if we would have the `raycast()` method return `Sequence[float, float], boolean`, but we're concerned about performance issues. I'll be adding the iterable return-type and compare the performance differences between the two to see how bad it would be.

Another concern @itzpr3d4t0r has is if we wanted to call `raycast()` over and over again but with the same exact parameters, it would be unnecessarily expensive performance-wise which is why we could do `raycast(Line)`, so that `Line` would cache the result. Although, I am not for this idea and I would rather see users implement a caching system themselves.

src_c/geometry.c
```diff
@@ -8,6 +8,106 @@
...
+            return RAISE(PyExc_TypeError, "the starting position requires a pair of floats");
+        }
+        if (!pg_TwoDoublesFromObj(args[1], &end_x, &end_y)) {
+            return RAISE(PyExc_TypeError, "the end position requires a pair of floats");
+        }
+    }
+    else if (nargs == 4) {
+        if (!PySequence_FAST_CHECK(args[3])) {
+            return RAISE(PyExc_TypeError, "argument of raycast() must be a sequence");
+        }
```
itzpr3d4t0r, line 33: This is not needed, you had a duplicated check here i just noticed.

novialriptide: Whoops.

itzpr3d4t0r:
I guess there's just the return objects matter to finalize and that's it.

novialriptide:
> I guess there's just the return objects matter to finalize and that's it.

Yup! Finally. We can discuss the return objects matter through Discord some other time, I've got a lot of things on my calendar this and next week.

geometry.pyi
```diff
@@ -194,3 +191,16 @@ class Polygon:
...
+    endpos: Coordinate,
+    colliders: List[Rect, Circle, Line],
+) -> Coordinate: ...
+@overload
+def raycast(
+    originpos: Coordinate,
+    angle: float,
+    max_dist: float,
+    colliders: List[Rect, Circle, Line],
+) -> Coordinate: ...
```
itzpr3d4t0r, line 206: Here you should use Sequence[Rect, Circle, Line]

novialriptide: Good catch

geometry.pyi
```diff
@@ -194,3 +191,16 @@ class Polygon:
 def regular_polygon(
     sides: int, center: Coordinate, radius: float, angle: float = 0
 ) -> Polygon: ...
+@overload
+def raycast(
+    originpos: Coordinate,
+    endpos: Coordinate,
+    colliders: Sequence[Rect, Circle, Line],
```
itzpr3d4t0r, line 198: Actually now that i think about it this should be 
```suggestion
    colliders: Sequence[Union[Rect, Circle, Line]],
```

geometry.pyi
```diff
@@ -194,3 +191,16 @@ class Polygon:
...
+) -> Coordinate: ...
+@overload
+def raycast(
+    originpos: Coordinate,
+    angle: float,
+    max_dist: float,
+    colliders: Sequence[Rect, Circle, Line],
```
itzpr3d4t0r, line 205: This one as well

novialriptide:
This pull request has been taken over by @Emc2356.

src_c/collisions.c
```diff
@@ -314,3 +314,114 @@ pgCollision_PolygonPoint(pgPolygonBase *poly, double x, double y)
...
+    double xc = circle->x;
+    double yc = circle->y;
+
+    double dx = x2 - x1;
+    double dy = y2 - y1;
+    double A = dx * dx + dy * dy;
+    double B = 2 * (dx * (x1 - xc) + dy * (y1 - yc));
+    double C = (x1 - xc) * (x1 - xc) + (y1 - yc) * (y1 - yc) - circle->r_sqr;
+
+    double descriminant = B * B - 4 * A * C;
```
itzpr3d4t0r, line 411: type: discriminant not descriminant

examples/raycast.py
```diff
@@ -49,21 +49,32 @@ def random_pos(width, height):
...
 ray_count = 360
 
+directions = [(math.cos(angle), math.sin(angle)) for x in range(ray_count)]
+
 while running:
+    for e in pygame.event.get():
+        if e.type == pygame.QUIT:
+            running = False
+            pygame.quit()
+            raise SystemExit
```
itzpr3d4t0r, line 61: Don't like this `raise SystemExit`

examples/raycast.py
```diff
@@ -49,21 +49,32 @@ def random_pos(width, height):
 
 running = True
 
-ray_endpoint = pygame.Vector2()
 ray_count = 360
 
+directions = [(math.cos(angle), math.sin(angle)) for x in range(ray_count)]
```
itzpr3d4t0r, line 54: Don't you see the issue here? you are basically doing calculations in python that could've easily been done in C with just the angle parameter, Same for line 72.

examples/raycast.py
```diff
@@ -49,21 +49,32 @@ def random_pos(width, height):
...
+    for e in pygame.event.get():
+        if e.type == pygame.QUIT:
+            running = False
+            pygame.quit()
+            raise SystemExit
+
     screen.fill((0, 0, 0))
 
     for x in range(ray_count):
         origin_pos = pygame.mouse.get_pos()
```
itzpr3d4t0r, line 66: this should be a single call outside the loop

src_c/geometry.c
itzpr3d4t0r: This must be resolved again, as @Emc2356 reintroduced it

src_c/geometry.c
```diff
@@ -8,6 +8,106 @@
...
+            return RAISE(PyExc_TypeError,
+                         "the origin requires a pair of floats");
+        }
+        double max_dist;
+        if (!pg_DoubleFromObj(args[2], &max_dist)) {
+            return RAISE(PyExc_TypeError, "invalid max distance");
+        }
+        if (max_dist < 0) {
+            max_dist = DBL_MAX;
+        }
```
itzpr3d4t0r, line 31: You should check and error out if the `max_dist == 0`.

src_c/geometry.c
```diff
@@ -8,6 +8,106 @@
...
+    PyObject **farr;
+    Py_ssize_t farr_length;
+    Py_ssize_t loop;
+    double max_t;
+    pgLineBase line;
+
+    if (nargs == 4) {
+        if (!pg_TwoDoublesFromObj(args[0], &line.x1, &line.y1)) {
+            return RAISE(PyExc_TypeError,
+                         "the origin requires a pair of floats");
```
itzpr3d4t0r, line 23: The error message is too generic and a bit misleading... I'd put something like: `"Invalid ray origin value, must be a pair of numeric values"`

src_c/geometry.c
```diff
@@ -8,6 +8,106 @@
...
+        }
+        double max_dist;
+        if (!pg_DoubleFromObj(args[2], &max_dist)) {
+            return RAISE(PyExc_TypeError, "invalid max distance");
```
itzpr3d4t0r, line 27: As an error message I'd put `"Invalid ray max distance threshold value, must be numeric"`

src_c/geometry.c
```diff
@@ -8,6 +8,106 @@
...
+        }
+        if (max_dist < 0) {
+            max_dist = DBL_MAX;
+        }
+        if (!pg_TwoDoublesFromObj(args[1], &line.x2, &line.y2)) {
+            double angle;
+            if (!pg_DoubleFromObj(args[1], &angle)) {
+                return RAISE(PyExc_TypeError,
+                             "expected a pair of floats or a single float");
```
itzpr3d4t0r, line 36: You don't expect a pair of floats from an angle parameter... Rephrase this error message to something like: `"Invalid ray angle value, must be numeric"`

src_c/geometry.c
```diff
@@ -8,6 +8,106 @@
...
+        double max_dist;
+        if (!pg_DoubleFromObj(args[2], &max_dist)) {
+            return RAISE(PyExc_TypeError, "invalid max distance");
+        }
+        if (max_dist < 0) {
+            max_dist = DBL_MAX;
+        }
+        if (!pg_TwoDoublesFromObj(args[1], &line.x2, &line.y2)) {
+            double angle;
+            if (!pg_DoubleFromObj(args[1], &angle)) {
```
itzpr3d4t0r, line 34: Instead of this nested tentatives of gettign a number or a pair of numbers (which can be cumputationally expensive) you could just check whether the `args[0]` is a number with `PyNumber_Check` and if so use `pg_DoubleFromObj` else directly try to use `pg_TwoDoublesFromObj`. in this way it's more error safe, more performant and easier to read.

src_c/geometry.c
```diff
@@ -8,6 +8,106 @@
...
+            }
+
+            if (max_dist != DBL_MAX) {
+                line.x2 = line.x1 - cos(angle * PI / 180) * max_dist;
+                line.y2 = line.y1 - sin(angle * PI / 180) * max_dist;
+            }
+            else {
+                line.x2 = line.x1 - cos(angle * PI / 180);
+                line.y2 = line.y1 - sin(angle * PI / 180);
+            }
```
itzpr3d4t0r, line 46: All of this you could put inside the if at line 29, but the thing is that you should extract the `max_dist `value AFTER the angle/end_pos value. in this way code is more compact and organized.

itzpr3d4t0r (changes requested):
This is just the first part of the review because i gtg, be ready for a second one after you made the changes.

src_c/collisions.c
```diff
@@ -314,3 +314,114 @@ pgCollision_PolygonPoint(pgPolygonBase *poly, double x, double y)
...
+    if (discriminant < 0) {
+        return 0;
+    }
+    double t = (-B - sqrt(discriminant)) / (2 * A);
+    if (t < 0) {
+        t = (-B + sqrt(discriminant)) / (2 * A);
+        if (t < 0) {
+            return 0;
+        }
+    }
```
itzpr3d4t0r, line 421: In the worst case scenario you are calculating the `sqrt(discriminant)` twice, which isn't ideal. you could store that in a variable and reuse it.

src_c/collisions.c
```diff
@@ -314,3 +314,114 @@ pgCollision_PolygonPoint(pgPolygonBase *poly, double x, double y)
...
+    double x2 = line->x2;
+    double y2 = line->y2;
+    double xc = circle->x;
+    double yc = circle->y;
+
+    double dx = x2 - x1;
+    double dy = y2 - y1;
+    double A = dx * dx + dy * dy;
+    double B = 2 * (dx * (x1 - xc) + dy * (y1 - yc));
+    double C = (x1 - xc) * (x1 - xc) + (y1 - yc) * (y1 - yc) - circle->r_sqr;
```
itzpr3d4t0r, line 409: you could store both `(x1 - xc)` and `(y1 - yc)` into separate values and calculate them once instead of three times

src_c/geometry.c
```diff
@@ -8,6 +8,106 @@
 
 #define PYGAMEAPI_GEOMETRY_NUMSLOTS 21
 
+static PyObject *
+pg_raycast(PyObject *_null, PyObject *const *args, Py_ssize_t nargs)
+{
+    PyObject **farr;
+    Py_ssize_t farr_length;
```
itzpr3d4t0r, line 15: I know i did it myself, but we should rename all `farr` thingies into more readable things, like i would suggest you rename `farr` -> `colliders` and `farr_length` -> `colliders_length`

src_c/geometry.c
```diff
@@ -8,6 +8,106 @@
...
+                                   &temp_t)) {
+                record_t = MIN(record_t, temp_t);
+            }
+        }
+        else if (pgRect_Check(farr[loop])) {
+            if (pgRaycast_LineRect(&line, &pgRect_AsRect(farr[loop]), max_t,
+                                   &temp_t)) {
+                record_t = MIN(record_t, temp_t);
+            }
+        }
```
itzpr3d4t0r, line 94: Not that it matters much, though it would shorten the code length a bit, you could store the `farr[loop]` into a `PyObject* obj`
```suggestion
        PyObject *obj = farr[loop];
        if (pgCircle_Check(obj)) {
            if (pgRaycast_LineCircle(&line, &pgCircle_AsCircle(obj),
                                     max_t, &temp_t)) {
                record_t = MIN(record_t, temp_t);
            }
        }
        else if (pgLine_Check(obj)) {
            if (pgRaycast_LineLine(&line, &pgLine_AsLine(obj), max_t,
                                   &temp_t)) {
                record_t = MIN(record_t, temp_t);
            }
        }
        else if (pgRect_Check(obj)) {
            if (pgRaycast_LineRect(&line, &pgRect_AsRect(obj), max_t,
                                   &temp_t)) {
                record_t = MIN(record_t, temp_t);
            }
        }
```

test/test_line.py
```diff
@@ -1,9 +1,8 @@
 import unittest
-from math import sqrt
 
 from pygame import Vector2, Vector3, Rect
 
-from geometry import Circle, Line
+from geometry import Circle, Line, raycast
```
itzpr3d4t0r, line 5: this should be removed

itzpr3d4t0r:
I think we should also abstract away the whole "get something that looks like a ray out of the params list" into a separate function Like `pgRay_FromFastcallArgs(args**, nargs-1, line*)` and `pgRay_FromObject(PyObject*, line*)`. This wouyld help drammatically reduce the future `multiraycast()` implementation

examples/raycast.py
```diff
@@ -49,21 +49,26 @@ def random_pos(width, height):
...
 
 while running:
+    for e in pygame.event.get():
+        if e.type == pygame.QUIT:
+            running = False
+
     screen.fill((0, 0, 0))
 
     for x in range(ray_count):
         origin_pos = pygame.mouse.get_pos()
```
itzpr3d4t0r, line 62: this should be a single call outside the loop

src_c/geometry.c
```diff
@@ -8,6 +8,150 @@
...
+
+        *max_t = max_dist / pgLine_Length(line);
+
+        return 1;
+    }
+    else {
+        PyErr_SetString(PyExc_TypeError, "Invalid number of arguments");
+        return 0;
+    }
+}
```
itzpr3d4t0r, line 87: Ok this came out exactly how i imagined it kudos!

src_c/geometry.c
```diff
@@ -8,6 +8,150 @@
...
+                PyExc_ValueError,
+                "Invalid ray max distance threshold value, must be numeric");
+            return 0;
+        }
+        if (max_dist < 0 || max_dist == DBL_MAX) {
+            *max_t = DBL_MAX;
+            return 1;
+        }
+        else if (max_dist == 0) {
+            PyErr_SetString(PyExc_ValueError, "max distance can not be 0");
```
itzpr3d4t0r, line 73: Maybe a slightly more precise error msg here?

src_c/geometry.c
```diff
@@ -8,6 +8,150 @@
...
+                return 0;
+            }
+
+            line->x2 = line->x1 - cos(angle * PI / 180);
+            line->y2 = line->y1 - sin(angle * PI / 180);
+        }
+        else if (!pg_TwoDoublesFromObj(args[1], &line->x2, &line->y2)) {
+            PyErr_SetString(PyExc_TypeError,
+                            "expected a pair of floats or a single float");
+            return 0;
```
itzpr3d4t0r, line 58: You don't expect a single float anymore, so you can change the error message now.

PR #128, opened 2022-10-19
Implemented Polygon.collidepoint()

itzpr3d4t0r:
One of the key features of shapes, allows you to do:

```Python
poly.collidepoint((x, y))
# or
poly.collidepoint(x, y)
```


src_c/polygon.c
```diff
@@ -460,9 +460,38 @@ pg_polygon_move_ip(pgPolygonObject *self, PyObject *const *args,
...
+        }
+    }
+    else if (nargs == 2) {
+        if (!pg_DoubleFromObj(args[0], &x) || !pg_DoubleFromObj(args[1], &y)) {
+            goto error;
+        }
+    }
+    else {
+        goto error;
+    }
```
Emc2356, line 481: so far we have used this code a lot of times, i think it is time to put it as a function in `geometry.c`/`geometry.h` and name it like _pgTwoDoublesFromFastcall

src_c/collisions.c
```diff
@@ -289,3 +289,29 @@ pgCollision_RectCircle(SDL_Rect *rect, pgCircleBase *circle)
...
+        double yj = poly->vertices[j * 2 + 1];
+
+        if (((yi > y) != (yj > y)) &&
+            (x < (xj - xi) * (y - yi) / (yj - yi) + xi)) {
+            collision = !collision;
+        }
+    }
+
+    return collision;
+}
```
Emc2356, line 317: you need a newline here

Emc2356 (approved):
overall looks great

src_c/polygon.c
itzpr3d4t0r: I think we should do that just not in this PR.

itzpr3d4t0r: I'll open another PR that adds that and applies that where necessary.

PR #129, opened 2022-10-23
geometry.regular_polygon() optimization

itzpr3d4t0r:
I've found a little optimization due to the fact that `pgPolygonNew2` utilizes `_pg_polygon_subtype_new2`. The latter has the peculiarity to allocate new memory, copy from the input one and also has to calculate the center of the polygon, therefore having to run trough a for loop for all the vertices.

With this optimization i basically skip new memory assigment (like in `_pg_polygon_subtype_new2_transfer`) by directly assigning the already available memory and i also skip the whole memcpy step and pymemfree. I also easily skip the "running trough all the verticies" as we already know where the center of the poly should be because it's the values we passed as params to the function. 

I've also swapped all ints for Py_ssize_t(s) as it's more scalable this way and makes this work on all platform configs.

With all these changes i've seen a 9-11% perf improvement.

Edit.
I implemented a more performant path for when the function is asked to create a poly with an even number of vertices. In that case we don't need to compute all the vertices but rather half of them and mirror the rest 180° apart. this reduces the number of sin/cos by half and the number of multiplications by half. I've seen performance uplifts of up to 2X for very big polys and around 33% for squares/esagons.

### DIFF
--- a/src_c/geometry.c
+++ b/src_c/geometry.c
@@ -12,7 +12,7 @@ static PyObject *
 geometry_regular_polygon(PyObject *_null, PyObject *const *args,
                          Py_ssize_t nargs)
 {
-    int sides;
+    Py_ssize_t sides;
     double radius;
     double angle = 0;
     double Cx, Cy;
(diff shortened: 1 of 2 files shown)


src_c/geometry.c
```diff
@@ -56,20 +56,29 @@ geometry_regular_polygon(PyObject *_null, PyObject *const *args,
...
 
-    return ret;
-}
+    if (!ret) {
+        PyMem_Free(vertices);
+        return NULL;
+    }
 
+    ret->polygon.vertices = vertices;
+    ret->polygon.verts_num = (Py_ssize_t)sides;
```
Emc2356, line 76: `sides` is already `Py_ssize_t`

Emc2356 (approved):
i really like this aproach, just that single thing and we are ok for a merge

Emc2356 (approved):
LGTM!

PR #135, opened 2022-10-27
Added pg_TwoDoublesFromFastcallArgs and swapped all duplicates

itzpr3d4t0r:
This should address a comment from @Emc2356 in #128. This PR swaps all occurrences of two doubles extraction from the fastcall args with a common inline function. I tested with and without the inline and i noticed that the inline keeps performance way closer to the original, though a touch (approx 1%) slower. This should make our code shorter and more readable and maintainable over time.


Emc2356 (approved):
the slowdown is minor, and if we need it this can become a geometry only thing and inline it for the slight performance increase, so i think this should be ok to merge

PR #138, opened 2022-10-31
implemented Line.at

Emc2356:


geometry.pyi
```diff
@@ -118,6 +118,7 @@ class Line(Sequence[float]):
     def move_ip(self, x: float, y: float) -> None: ...
     @overload
     def move_ip(self, move_by: Coordinate) -> None: ...
+    def at(self, t: float) -> Tuple[float, float]: ...
```
itzpr3d4t0r, line 121: Given that this is basically a lerp, wouldn't something like `weight` or `fac` be better suited as "t"?

test/test_line.py
```diff
@@ -905,6 +905,22 @@ def test_slope_getter(self):
...
+        for value in invalid_types:
+            with self.assertRaises(TypeError):
+                line.at(value)
+
+        self.assertEqual(line.at(0.5), (0.5, 0.5))
+        self.assertEqual(line.at(1), (1, 1))
+        self.assertEqual(line.at(0), (0, 0))
+        self.assertEqual(line.at(-1), (-1, -1))
+        self.assertEqual(line.at(2), (2, 2))
+
```
itzpr3d4t0r, line 923: You should add some checks that ensure that the function always returns a Tuple type

src_c/line.c
```diff
@@ -430,6 +437,21 @@ pg_line_move_ip(pgLineObject *self, PyObject *const *args, Py_ssize_t nargs)
...
 }
 
+static PyObject *
+pg_line_at(pgLineObject *self, PyObject *obj)
+{
+    double t;
+    double x, y;
+
+    if (!pg_DoubleFromObj(obj, &t)) {
+        return RAISE(PyExc_TypeError, "excpected a numeric value");
```
itzpr3d4t0r, line 447: better error message? also there's a typo in there

PR #140, opened 2022-10-31
`geometry.multiraycast()`

itzpr3d4t0r:


src_c/geometry.c
```diff
@@ -243,9 +243,171 @@ geometry_regular_polygon(PyObject *_null, PyObject *const *args,
...
+        if (record_t == max_t) {
+            if (include_misses) {
+                Py_INCREF(Py_None);
+                PyList_SET_ITEM(ret, i, Py_None);
+            }
+        }
+        else {
+            PyObject *point =
+                pg_TupleFromDoublePair(ray.x1 + (ray.x2 - ray.x1) * record_t,
+                                       ray.y1 + (ray.y2 - ray.y1) * record_t);
```
Emc2356, line 384: i would suggest to maybe use `pgLine_At`?

novialriptide:
For now, I'm putting the dont-merge tag since `include_miss=False` would return an array of collision points whose length is unequal to the number of rays that are inputted into the function. (Let me know if my explanation is a bit of a headache to understand lol).

itzpr3d4t0r:
> For now, I'm putting the dont-merge tag since `include_miss=False` would return an array of collision points whose length is unequal to the number of rays that are inputted into the function. (Let me know if my explanation is a bit of a headache to understand lol).

That was inteded as an optimization, but i just removed it because it wasn't worth the performance and complexity

PR #160, opened 2022-12-19
Added more jobs for each OS, covering CP3.7-CP3.11

itzpr3d4t0r:
This pull request adds GitHub Actions jobs for Python 3.7 to 3.11 on the latest versions of macOS, Ubuntu, and Windows. It also fixes an issue with the CPython 3.7 jobs, where the math.dist function was not present in the math module at that time. However, there are currently some **known issues** with these jobs:

- On macOS, building with CPython 3.7-3.11 and pygame 2.1.2 is not possible. It only works when using pygame 2.1.3dev8. This is believed to be related to the inclusion of the immintrin.h library.

- On Windows, building with CPython 3.11 and pygame 2.1.2 is also not possible. It only works when using pygame 2.1.3dev8.

This means we still need to find the cause of the issue and resolve it in a separate PR.


novialriptide (approved):
LGTM

PR #168, opened 2023-01-11
Polygon `as_segments()`

### LINKED ISSUE #155: Add as_segments() function to Polygon
This method would return a list of Lines that represents the polygon as a series of connected line segments.

Here is an example of how it could be used in Python:

```Python
from geometry import Polygon, Line

polygon = Polygon([(0, 0), (1, 0), (1, 1), (0, 1)])

# Get a list of the sides of the polygon
segments = polygon.as_segments()

# segments is a list of Line objects with the following positions:
# [Line((0, 0), (1, 0)), Line((1, 0), (1, 1)), Line((1, 1), (0, 1)), Line((0, 1), (0, 0))]

# Print the length of each side
for segment in segments:
    print(segment.length)

```
The output of this program would be:
```
1.0
1.0
1.0
1.0
```

ScriptLineStudios:
Closes #155


docs/polygon.rst
```diff
@@ -148,6 +148,15 @@ Polygon Methods
 
       .. ## Polygon.collidepoint ##
 
+    .. method:: as_segments
+
+        | :sl:`returns the line segments of the polygon`
+        | :sg:`as_segments() -> list`
```
itzpr3d4t0r, line 154: Well thanks for contributing again! I'll leave some reviews

itzpr3d4t0r, line 154:
```suggestion
        | :sg:`as_segments() -> list[pygame.Rect]`
```

src_c/polygon.c
```diff
@@ -468,6 +468,31 @@ pg_polygon_move(pgPolygonObject *self, PyObject *const *args, Py_ssize_t nargs)
     return tmp;
 }
 
+static PyObject *
+pg_polygon_as_segments(pgPolygonObject *self) {
```
itzpr3d4t0r, line 472: `METH_NOARGS` functions should have a second parameter that's `_null`
```suggestion
pg_polygon_as_segments(pgPolygonObject *self, PyObject *_null) {
```

src_c/polygon.c
```diff
@@ -468,6 +468,31 @@ pg_polygon_move(pgPolygonObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+    PyObject *list = PyList_New(0);
```
itzpr3d4t0r, line 473: Given that you know how many vertices you have in a polygon you can preallocate a list of size `verts_num` with `PyObject *list = PyList_New(self->polygon.verts_num);`. Later on for appending to a preallocated list you must use the `PyList_SET_ITEM` macro.

src_c/polygon.c
```diff
@@ -468,6 +468,31 @@ pg_polygon_move(pgPolygonObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+            next_vertex_y = self->polygon.vertices[1];
+        }
+        else {
+            next_vertex_x = self->polygon.vertices[i + 2];
+            next_vertex_y = self->polygon.vertices[i + 3];
+        }
+
+        PyList_Append(list, pgLine_New4(
+            self->polygon.vertices[i], self->polygon.vertices[i + 1], 
+            next_vertex_x, next_vertex_y)
```
itzpr3d4t0r, line 489: This has two problems:
1. You should use `PyList_SET_ITEM`
2. You are not making sure that pgLine_New4() returns a valid object and not NULL

geometry.pyi
```diff
@@ -208,6 +208,8 @@ class Polygon:
     @overload
     def move(self, move_by: Coordinate) -> Polygon: ...
     @overload
+    def as_segments(self) -> List[Polygon]: ...
```
itzpr3d4t0r, line 211:
```suggestion
    def as_segments(self) -> List[Line]: ...
```

src_c/polygon.c
```diff
@@ -468,6 +468,34 @@ pg_polygon_move(pgPolygonObject *self, PyObject *const *args, Py_ssize_t nargs)
     return tmp;
 }
 
+static PyObject *
+pg_polygon_as_segments(pgPolygonObject *self, PyObject *_null) {
+    double *verts = self->polygon.vertices;
+    Py_ssize_t vert_num = self->polygon.verts_num;
+    PyObject *list = PyList_New(vert_num);
```
itzpr3d4t0r, line 475: You should check if list is NULL and exit and return NULL if so.

src_c/polygon.c
```diff
@@ -468,6 +468,34 @@ pg_polygon_move(pgPolygonObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+            next_vertex_x = verts[i + 2];
+            next_vertex_y = verts[i + 3];
+        }
+
+        PyObject *line = pgLine_New4(verts[i], verts[i + 1], next_vertex_x, next_vertex_y);
+        if (!line) {
+            return NULL;
+        }
+        PyList_SET_ITEM(list, i / 2, line);
+    }
```
itzpr3d4t0r, line 494: To be honest you could make this much simpler by doing this:
```suggestion
    for (i = 0; i < verts_num; i++) {
        Py_ssize_t i2 = i * 2;
        PyObject *line = pgLine_New4(vertices[i2], vertices[i2 + 1],
                                        vertices[(i2 + 2) % (verts_num*2)],
                                        vertices[(i2 + 3) % (verts_num*2)]);
        if (!line) {
            Py_DECREF(list);
            return NULL;
        }
        PyList_SET_ITEM(list, i, line);
    }
```

src_c/polygon.c
```diff
@@ -468,6 +468,34 @@ pg_polygon_move(pgPolygonObject *self, PyObject *const *args, Py_ssize_t nargs)
     return tmp;
 }
 
+static PyObject *
+pg_polygon_as_segments(pgPolygonObject *self, PyObject *_null) {
+    double *verts = self->polygon.vertices;
+    Py_ssize_t vert_num = self->polygon.verts_num;
```
itzpr3d4t0r, line 474: I'd rename this one to verts_num
```suggestion
    Py_ssize_t verts_num = self->polygon.verts_num;
```

src_c/polygon.c
itzpr3d4t0r: you could also put that verts_num*2 into a separate variable

src_c/polygon.c
```diff
@@ -468,6 +468,34 @@ pg_polygon_move(pgPolygonObject *self, PyObject *const *args, Py_ssize_t nargs)
...
 
+static PyObject *
+pg_polygon_as_segments(pgPolygonObject *self, PyObject *_null) {
+    double *vertices = self->polygon.vertices;
+    Py_ssize_t verts_num = self->polygon.verts_num;
+    Py_ssize_t verts_num_double = verts_num * 2;
+    
+    PyObject *list = PyList_New(verts_num);
+    if (!list) {
+        Py_DECREF(list);
```
itzpr3d4t0r, line 479: There's no need to DECREF a NULL list

test/test_polygon.py
```diff
@@ -449,6 +449,33 @@ def test__repr__(self):
...
+                Line((123.23, 35.6), (56.4, 87.45)),
+                Line((56.4, 87.45), (43.1, 12.3)),
+                Line((43.1, 12.3), (123.23, 35.6)),
+            ],
+        )
+        poly = Polygon([[1, 2], [3, 4], [5, 6]])
+        self.assertEqual(
+            poly.as_segments(),
+            [Line((1, 2), (3, 4)), Line((3, 4), (5, 6)), Line((5, 6), (1, 2))],
+        )
```
itzpr3d4t0r, line 477: You should add more tests for invalid argument types and argument number.

itzpr3d4t0r:
I would be thankful if you could add a couple benchmarks as well in the benchmarks directory, though that's not strictly necessary, i can add them later.

geometry.pyi
```diff
@@ -208,6 +208,8 @@ class Polygon:
     @overload
     def move(self, move_by: Coordinate) -> Polygon: ...
     @overload
+    def as_segments(self) -> List[Line]: ...
```
itzpr3d4t0r, line 211: you don't need @overload here, there's just one way to call this function

test/test_polygon.py
itzpr3d4t0r: Basically you should make sure that this is actually a no arguments method now and in the future

PR #185, opened 2023-01-13
`geometry.rect_to_polygon()`

### LINKED ISSUE #149: Pygame RectStyle to Polygon function
It would be pretty useful especially for having a quick way to convert a rect to a polygon for then operating transformations such as rotations that aren't possible on a pygame rect.

I think the function should be METH_FASTCALL in tune with the pygame implementation, where functions that accept rectangles also accept RectSyle inputs.

I believe it is possible to implement this same function in two ways, namely as part of the Polygon class or as a standalone function in the general geometry module. In the first case it would be a matter of having a function of the type `polygon.from_rect()` or something similar, while in the second case something like `geometry.polygon_fromrect()`.

Personally, I preferred the latter since having a function that can be called from any variable of type Polygon could misunderstand its use, which is much more generic and not limited to being a member function of the Polygon class, and also there could be cases such as tomato.from_rect() where the tomato variable is a Polygon object, but that is not immediately readable and could confuse the inattentive user. Not to mention homogeneous sets such as lists/sets and tuples.

Ideally it would be best to implement this as a Rect function so that one could simply do `my_rect.as_polygon()`, but that's not possible atm.

itzpr3d4t0r:
Closes #149.

This PR adds:
- The function `geometry.rect_to_polygon()`
- A new internal C function called `pgPolygon_New4`, which creates a new polygon using the provided vertices, number of vertices, and center.
- A new test file `test_geometry.py` to test all non-raycasting geometry functions.
- docs


src_c/geometry.c
```diff
@@ -244,10 +244,51 @@ geometry_regular_polygon(PyObject *_null, PyObject *const *args,
...
+            return NULL;
+        }
+        return RAISE(PyExc_TypeError, "rect must be a valid Rect object");
+    }
+
+    _normalize_rect(tmp);
+
+    double vertices[] = {tmp->x, tmp->y,          tmp->x + tmp->w,
+                         tmp->y, tmp->x + tmp->w, tmp->y + tmp->h,
+                         tmp->x, tmp->y + tmp->h};
```
itzpr3d4t0r, line 277: I don't know why but clang format made this strange arrangement and considers it valid.

Emc2356 (approved):

PR #187, opened 2023-01-14
added Polygon.is_convex method with docs and tests

Emc2356:
created a method for the polygon class that checks if the polygon is convex or concave
this also has docs and tests


docs/polygon.rst
```diff
@@ -200,6 +200,19 @@ Polygon Methods
 
       .. ## Polygon.rotate_ip ##
 
+    .. method:: is_convex
+
+        | :sl:`checks if the polygon is convex or concave`
```
itzpr3d4t0r, line 205: maybe this is a bit better
```suggestion
        | :sl:`checks whether the polygon is convex or concave`
```

docs/polygon.rst
```diff
@@ -200,6 +200,19 @@ Polygon Methods
...
+        | :sg:`is_convex() -> bool`
+
+        Checks whether the polygon is convex or concave.
+
+        .. note::
+            Keep in mind that the more vertices the polygon has, the more CPU time it will
+            take to move it.
+
+      .. ## Polygon.move ##
```
itzpr3d4t0r, line 214:
```suggestion
      .. ## Polygon.is_convex ##
```

docs/polygon.rst
```diff
@@ -200,6 +200,19 @@ Polygon Methods
...
+    .. method:: is_convex
+
+        | :sl:`checks if the polygon is convex or concave`
+        | :sg:`is_convex() -> bool`
+
+        Checks whether the polygon is convex or concave.
+
+        .. note::
+            Keep in mind that the more vertices the polygon has, the more CPU time it will
+            take to move it.
```
itzpr3d4t0r, line 212:
```suggestion
            take to check if it's convex.
```

src_c/polygon.c
```diff
@@ -908,13 +908,98 @@ pg_polygon_rotate_ip(pgPolygonObject *self, PyObject *arg)
...
+{
+    // Stores direction of cross product
+    // of previous traversed edges
+    double prev = 0;
+
+    // Stores direction of cross product
+    // of current traversed edges
+    double curr = 0;
+
+    double *_memory = PyMem_New(double, 6);
```
itzpr3d4t0r, line 948: You don't need memory on the heap in this case

src_c/polygon.c
```diff
@@ -908,13 +908,98 @@ pg_polygon_rotate_ip(pgPolygonObject *self, PyObject *arg)
...
+    }
+    PyMem_Free(_memory);
+    return 1;
+}
+
+static PyObject *
+pg_polygon_is_convex(pgPolygonObject *self, PyObject *_null)
+{
+    int result = _pg_polygon_is_convex_helper((&self->polygon)->vertices,
+                                              (&self->polygon)->verts_num * 2);
```
itzpr3d4t0r, line 990: This has some unnecessary complexity. I would make `_pg_polygon_is_convex_helper` accept a `pgPolygonBase *` and then get the needed attributes inside that function. In this way you could reduce the python function to a single line.

itzpr3d4t0r: also maybe rename `_pg_polygon_is_convex_helper` to something like `_is_convex_helper`.

test/test_polygon.py
```diff
@@ -1035,6 +1035,16 @@ def test_polygon___new__(self):
...
 
+    def test_is_convex_meth(self):
+        p1 = Polygon((0, 0), (0, 1), (1, 1), (1, 0))
+        p2 = Polygon((0, 10), (5, 5), (10, 10), (10, 0), (0, 0))
+
+        with self.assertRaises(TypeError):
+            p1.is_convex(1)
+
+        self.assertTrue(p1.is_convex())
+        self.assertFalse(p2.is_convex())
```
itzpr3d4t0r, line 1046: You should add more tests for invalid arguments. They should check that this function is indeed a METH_NOARGS one.

src_c/polygon.c
```diff
@@ -908,13 +908,98 @@ pg_polygon_rotate_ip(pgPolygonObject *self, PyObject *arg)
...
+
+    // Return cross product
+    return X1 * Y2 - Y1 * X2;
+}
+
+/*
+ * https://www.geeksforgeeks.org/check-if-given-polygon-is-a-convex-polygon-or-not/
+ */
+static int
+_pg_polygon_is_convex_helper(double *verticies, int count)
```
itzpr3d4t0r, line 938: More comments about this function in general later

itzpr3d4t0r (changes requested):

src_c/polygon.c
```diff
@@ -908,13 +908,99 @@ pg_polygon_rotate_ip(pgPolygonObject *self, PyObject *arg)
...
+    // Stores direction of cross product
+    // of current traversed edges
+    double curr = 0;
+
+    double A[2];
+    double B[2];
+    double C[2];
+
+    // Traverse the array
+    Py_ssize_t i1;
```
itzpr3d4t0r, line 958: maybe just call this i

src_c/polygon.c
```diff
@@ -908,13 +908,69 @@ pg_polygon_rotate_ip(pgPolygonObject *self, PyObject *arg)
...
+
+    return 1;
+}
+
+static PyObject *
+pg_polygon_is_convex(pgPolygonObject *self, PyObject *_null)
+{
+    int result = _pg_polygon_is_convex_helper(&self->polygon);
+
+    return PyBool_FromLong(result);
```
itzpr3d4t0r, line 963: at this point put this on a single line

src_c/polygon.c
```diff
@@ -908,13 +908,69 @@ pg_polygon_rotate_ip(pgPolygonObject *self, PyObject *arg)
...
+{
+    Py_ssize_t i, i0, i1, i2;
+    Py_ssize_t verts_num = poly->verts_num;
+    Py_ssize_t count = 2 * verts_num;
+    double *vertices = poly->vertices;
+
+    /* A polygon is convex if and only if the cross products of all the
+     * adjacent edges are all of the same sign.
+     */
+    int sign = 0;
```
itzpr3d4t0r, line 927: The comment here could pe put before the variables and the sing variable closer to the others

src_c/polygon.c
```diff
@@ -908,13 +908,69 @@ pg_polygon_rotate_ip(pgPolygonObject *self, PyObject *arg)
...
+            /* The polygon is not convex if any two edges are colinear. */
+            return 0;
+        }
+        else if (sign == 0) {
+            sign = cross < 0 ? -1 : 1;
+        }
+        else if ((sign == -1 && cross > 0) || (sign == 1 && cross < 0)) {
+            /* The polygon is not convex if the cross products of any
+             * two adjacent edges are of different signs.
+             */
```
itzpr3d4t0r, line 950: don't like this end of comment on a separate line

itzpr3d4t0r (approved):

PR #203, opened 2023-02-18
Implement runtime check for AVX2 SIMD functionality in collisions

### LINKED ISSUE #202: Python crashes when running the unittests on Linux Mint
# Environment

 Linux Mint 21.1 on x86_64 AMD A8-7410 APU

# Reproduction 

 Running the unittest by `python3 -m unittest` in the repository directory

# Output

```
pygame-ce 2.1.4.dev1 (SDL 2.0.20, Python 3.10.6)
...............................................................................................................Illegal instruction (core dumped)
```

 This creates an illegal instruction segfault and crashes Python completely. I suspect, this has to do with it using a kind of SIMD vector extension that's not supported by my processor.

avaxar:
This PR implements a runtime check on whether the AVX2 instruction set is present or not in the CPU. Compile-time macros like `__AVX2__` only flags whether the compiler support generating code using AVX2, but not whether the running system would be able to execute it. If a non-AVX2-supporting processor runs an AVX2-using code, the program will generate a segfault over an unrecognized instruction/opcode, as what I intend to fix #202. Because pygame-geometry doesn't use SDL2, which means not being able to use the straightforward `SDL_HasAVX2` function, I had to make my own function to check it.

### DIFF
--- a/src_c/collisions.c
+++ b/src_c/collisions.c
@@ -201,8 +201,10 @@ pgIntersection_LineRect(pgLineBase *line, SDL_Rect *rect, double *X, double *Y,
                         double *T)
 {
 #if AVX2_IS_SUPPORTED
-    return pgIntersection_LineRect_avx2(line, rect, X, Y, T);
-#else
+    if (pg_HasAVX2())
+        return pgIntersection_LineRect_avx2(line, rect, X, Y, T);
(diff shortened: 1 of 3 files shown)


novialriptide (approved):
LGTM.

PR #204, opened 2023-03-12
created Polygon.area attribute

Emc2356:


itzpr3d4t0r:
This is still missing attr docs in polygon.rst

PR #226, opened 2023-08-09
Update CONTRIBUTING.md

gresm:
The repository was moved, so the links should be updated.


itzpr3d4t0r (approved):
Thanks for the PR! yeah that got definitely overlooked.

PR #229, opened 2023-12-28
rename 'collideswith' to 'collide'

### LINKED ISSUE #223: Rename "collideswith" function to "collide"
I suggest this change as it has two benefits:

1. It would make people's code shorter and a bit more readable since the function name isn't composed of two words without space
2. It would align more with the other collision functions which are in the form `.collide*` and not `.collideswith*`

LaggyWolf:
This change resolves #223 by renaming all 'collideswith' functions to 'collide'.


itzpr3d4t0r:
Thanks for the PR, But I'm not completely sure everyone agrees with this change... we shall see.

PR #242, opened 2025-12-13
Redirect dead links

andrewhong04:


andrewhong04:
Windows could not find a suitable runtime. "No suitable Python runtime found"

Merging.

